PR #138, opened 2023-04-05
Fix EditListingDetailsForm default prop

SariSaar:
EditListingDetailsForm: set listingFieldsConfig default prop to empty array to fix a bug with initial null value not getting a default in props destructuring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2023-XX-XX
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2023-XX-XX
 
+- [fix] EditListingDetailsForm: set listingFieldsConfig default prop to empty array.
```
SariSaar, line 17:
```suggestion
- [fix] EditListingDetailsForm: set listingFieldsConfig default prop to empty array to fix a bug with initial null value not getting a default in props destructuring.
```

PR #141, opened 2023-04-14
Fix layout-breaking condition in purchase templates

SariSaar:
Move conditions for delivery method to parallel the else block in purchase email templates, so that the table layout works with all delivery method options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2023-XX-XX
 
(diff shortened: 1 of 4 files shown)


Gnito (commented):
I have not watched closely these email templates but it looks good to me.
Maybe worth checking that Janne has seen these changes.

In addition: could you add a new entry to [CHANGELOG.md](https://github.com/sharetribe/web-template/blob/main/CHANGELOG.md#upcoming-version-2023-xx-xx)?

Gnito (approved):

PR #153, opened 2023-04-24
Fix booking end date in transaction breakdown for day-based bookings

bladealslayer:


Gnito (approved):
LGTM

Add changelog entry

PR #162, opened 2023-05-15
Add hosted footer functionality

SariSaar:
- Create FooterContainer
- Add new section for footer
- Add new blocks for footer and social media link
- Add new primitives for SocialMediaLink and Text
- Replace old Footer component with dynamic FooterContainer
- Remove hard-coded Footer component
- Add code splitting to default content page PageBuilder imports


src/containers/FooterContainer/FooterContainer.js
```diff
@@ -0,0 +1,24 @@
...
+const SectionBuilder = loadable(() => import('../PageBuilder/PageBuilder'), {
+  resolveComponent: components => components.SectionBuilder,
+});
+
+const FooterContainer = () => {
+  const { footer } = useConfiguration();
+
+  // The footer asset does not specify sectionId or sectionType. However, the SectionBuilder
+  // expects sectionId and sectionType in order to identify the section. I suggest that we
+  // add those attributes here before passing the asset to SectionBuilder.
```
Gnito, line 14: Hmm. Don't leave "I suggest" to code comments.

We could ask for a change to the footer asset schema or just describe that we add those attributes here.

However, it would feel weird to ask footer id from the operator in content/footer.json, which leads to the situation that the id would need to be hard-coded somewhere. So, I agree that this is as good a place as the asset schema file for this (rather odd) "section" config.

src/containers/FooterContainer/footerasset.json
```diff
@@ -0,0 +1,122 @@
+{
```
Gnito, line 1: If I understood correctly, this is currently an unused file?

src/containers/LandingPage/FallbackPage.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
-import PageBuilder from '../PageBuilder/PageBuilder';
+import loadable from '@loadable/component';
+
 import css from './FallbackPage.module.css';
 
+const PageBuilder = loadable(() => import('../../containers/PageBuilder/PageBuilder'));
```
Gnito, line 6:
```suggestion
const PageBuilder = loadable(() => import(/* webpackChunkName: "PageBuilder" */ '../../containers/PageBuilder/PageBuilder'));
```

https://webpack.js.org/api/module-methods/#magic-comments

src/containers/PageBuilder/BlockBuilder/BlockSocialMediaLink/BlockSocialMediaLink.js
```diff
@@ -0,0 +1,55 @@
...
+  mediaClassName: string,
+  textClassName: string,
+  ctaButtonClass: string,
+  title: object,
+  text: object,
+  callToAction: object,
+  media: object,
+  responsiveImageSizes: string,
+  options: propTypeOption,
+};
```
Gnito, line 53: You only extract these props: blockId, className, rootClassName, link
Therefore, I guess the other propTypes could be removed.

However, you should pass the **_options_** concept to `<Field>`. That makes this whole PageBuilder setup extendable.

src/containers/PageBuilder/Field/Field.helpers.js
```diff
@@ -41,6 +41,14 @@ export const exposeLinkProps = data => {
   return cleanUrl ? { children: linkText, href: cleanUrl } : {};
 };
 
+export const exposeSocialMediaProps = data => {
+  const { platform, url } = data;
+  const hasCorrectProps = typeof url === 'string' && url.length > 0;
+  const cleanUrl = hasCorrectProps ? sanitizeUrl(url) : null;
+  // TODO: Figure out how to validate platform!
```
Gnito, line 48: One simple way is to do something like this:
`const validPlatform = ['facebook', 'twitter'].includes(platform) ? platform : null;`

I.e. explicitly add all the supported platforms to the array.

src/containers/PageBuilder/Field/Field.js
```diff
@@ -211,6 +217,12 @@ const propTypeLink = shape({
   href: string.isRequired,
 });
 
+const propTypeSocialMediaLink = shape({
+  fieldType: oneOf(['socialMediaLink']).isRequired,
+  platform: string,
```
Gnito, line 222: Why this is not required?

src/containers/PageBuilder/Primitives/Link/SocialMediaLink.js
```diff
@@ -0,0 +1,62 @@
...
+SocialMediaLink.defaultProps = {
+  rootClassName: null,
+  className: null,
+};
+
+SocialMediaLink.propTypes = {
+  rootClassName: string,
+  className: string,
+  children: node.isRequired,
+  href: string.isRequired,
```
Gnito, line 61: Add "title" too with defaultProp as null. Some code editors pick these propTypes and offer them as guide for possible props when this component is used in JSX.

src/containers/PageBuilder/Primitives/Text/Text.js
```diff
@@ -0,0 +1,27 @@
...
+
+Text.displayName = 'Text';
+
+Text.defaultProps = {
+  rootClassName: null,
+  className: null,
+  children: null,
+};
+
+Text.propTypes = {
```
Gnito, line 23: Add "as" with default props: 'span',

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
```diff
@@ -0,0 +1,134 @@
...
+            <Field data={slogan} />
+          </div>
+          <div className={css.socialInfo}>
+            <div className={css.icons}>
+              {showSocialMediaLinks ? <BlockBuilder blocks={linksWithBlockId} /> : null}
+            </div>
+            <Field data={copyright} />
+          </div>
+          <div className={classNames(css.grid, getGridCss(numberOfColumns))}>
+            <BlockBuilder blocks={blocks} />
```
Gnito, line 87: Pass "options" to BlockBuilder (to make it consistently easy to extend blocks if someone needs to).

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
```diff
@@ -0,0 +1,134 @@
...
+        <div className={classNames(css.content, getContentCss(numberOfColumns))}>
+          <div>
+            <LinkedLogo className={css.logoLink} />
+          </div>
+          <div className={css.detailsInfo}>
+            <Field data={slogan} />
+          </div>
+          <div className={css.socialInfo}>
+            <div className={css.icons}>
+              {showSocialMediaLinks ? <BlockBuilder blocks={linksWithBlockId} /> : null}
```
Gnito, line 82: Pass "options" to BlockBuilder (to make it consistently easy to extend blocks if someone needs to).

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
```diff
@@ -0,0 +1,134 @@
...
+    ctaButton: string,
+  }),
+  numberOfColumns: number,
+  title: object,
+  description: object,
+  appearance: object,
+  callToAction: object,
+  blocks: arrayOf(object),
+  isInsideContainer: bool,
+  options: propTypeOption,
```
Gnito, line 131: These are outdated.

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
```diff
@@ -0,0 +1,185 @@
+@import '../../../../styles/customMediaQueries.css';
+
+.footer {
+  /* padding: 32px;
+  max-width: calc(var(--contentMaxWidth) + 72px);
+  margin: 0 auto; */
```
Gnito, line 6: What happens here?

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
```diff
@@ -0,0 +1,185 @@
...
+  padding: 32px;
+}
+
+.footer ul {
+  list-style-type: none;
+  padding-inline-start: 0px;
+}
+
+.linkList {
+  list-style-type: none;
```
Gnito, line 21: Add something about Footer to SectionBuilder/README.md

There are some special rules like this one. I.e. "list-item" doesn't look like a default bullet point list, because the expected use case is to create a list of links with that particular markdown syntax.

Gnito: Have you tested other markdown syntax in the footer - does it look OK?

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
```diff
@@ -0,0 +1,185 @@
...
+
+.gridCol2,
+.gridCol3,
+.gridCol4 {
+  grid-template-columns: 1fr 1fr;
+}
+
+.detailsInfo {
+  grid-area: details;
+  & span {
```
Gnito, line 57: If you can avoid it, don't use global elements. That leads to "specificity hell" in the long run. (In this context, it's not so bad, because you have used CSS Modules as a wrapper - but as a rule of thumb: instead of styling global elements, create classes, whenever you can.)

In this case, you should be able to just create a new class `.slogan` (no need to nest it)
and then pass the class name to `<Field data={slogan} className={css.slogan} />`

The reason why we style elements in PageBuilder is Markdown syntax, which creates plain HTML elements under the hood. (It's not as straightforward to pass custom classes to those elements.)

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
```diff
@@ -0,0 +1,185 @@
...
+  grid-area: logo;
+}
+
+.socialInfo {
+  grid-area: links;
+  & span {
+    font-size: 12px;
+    line-height: 16px;
+    color: #6e6e73;
+  }
```
Gnito, line 72: This could also be a class: `.copyright`.

This is actually pretty error-prone: it means that any span (however deeply nested) will get these rules.

You also use BlockBuilder inside the same .socialInfo div. So, when someone customizes social media links in the future, they might add <span> element there too. In that kind of scenario, it's usually quite frustrating to debug why some specific rules are applied and how to take them off without breaking something else.

```
          <div className={css.socialInfo}>
            <div className={css.icons}>
              {showSocialMediaLinks ? <BlockBuilder blocks={linksWithBlockId} /> : null}
            </div>
            <Field data={copyright} />
          </div>
```

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
```diff
@@ -0,0 +1,185 @@
...
+  font-size: 28px;
+  margin: 15px 15px 0px 0px;
+  padding: 10px 15px;
+  background-color: pink;
+}
+
+.logoLink {
+  padding: 0px;
+  height: 64px;
+}
```
Gnito, line 185: It's probably better to move these above the Media query breakpoints.

src/containers/LandingPage/FallbackPage.js
Gnito: Add this also to other dynamic imports (to help customizers pick this naming feature).

src/containers/FooterContainer/FooterContainer.js
SariSaar: Comment reworded 👍

src/containers/FooterContainer/footerasset.json
SariSaar: Yep, removed ✅

src/containers/LandingPage/__snapshots__/LandingPage.test.js.snap
```diff
@@ -1,553 +1,3 @@
...
-              </div>
-            </div>
-          </div>
-        </div>
-      </div>
-    </div>
-  </div>
-</div>
-`;
+exports[`LandingPage matches snapshot 1`] = `null`;
```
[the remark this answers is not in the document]
SariSaar: Added TODO item on LandingPage.test.js and reverted the test update.

src/containers/PageBuilder/Primitives/Link/SocialMediaLink.js
SariSaar: ✅

src/containers/PageBuilder/Primitives/Text/Text.js
SariSaar: ✅

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
SariSaar: Obsolete, cleaned up

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
SariSaar: Thanks for the explanation, fixed!

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
SariSaar: Thanks, fixed!

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
SariSaar: Fixed 👍

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
SariSaar: ✅

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.js
SariSaar: ✅

src/containers/PageBuilder/Field/Field.helpers.js
SariSaar: Exported a `supportedPlatforms` array from SocialMediaLink for this purpose and added a validation that uses that array here, thanks for the suggestion!

src/containers/PageBuilder/SectionBuilder/SectionFooter/SectionFooter.module.css
SariSaar: I tested markdown syntax in the footer, and bolding, italics and block quotes work. Heading levels work apart from font size, i.e. headings get bolding and added margin but stay the same size.

How detailed should the README.md explanation be on SectionFooter?

Gnito:
Cherry-picked comments to a new PR #163 
Closing this one.

PR #195, opened 2023-08-03
Inquiry process and changes for v3.0.0

Gnito:
**This branch has evolved to be the _v3_ development branch**

This adds support for a new process: **_default-inquiry_**. That process is super simple and meant for situations where payments are not supported or where the process needs to be separately developed (i.e. an MVP-style process to test some idea)

Essentially, **_default-inquiry_** process just enables free messaging between the customer and provider. Any payments etc. need to happen face-to-face or somehow outside of the marketplace.

**Notes:** 
- The initial inquiry message is saved to the protected data of the transaction.
- The process has 1 transition
- Price can be configured on/off 
  - It is then reflected on ListingCard, ListingPage, EditListingWizard, ManageListingCard, and CheckoutPage.

## Major changes:
- **This will change the CheckoutPage** quite much¹. 
- This also causes a major version update to the template. (v3.0.0)
- SearchMap: **SearchMapPriceLabel is empty** if the listing is working with an inquiry listing type that doesn't display a price.
  - We might introduce a separate label component later.
- **Remove en.json** file (Biketribe-themed translations) and 
- **Rename defaultMicrocopy.json** (with generic copy-texts) to be the new **en.json** file².
- **_server/index.js_ and _server/api/login-as.js_ use _getSdk_ function from api-util/sdk.js**
- ModalMissingReminder: StripeAccountReminder is removed.
- Brand image (for authentication page etc.) is not going to use default from the app but uses hosted image only.

¹ Read more about how the CheckoutPage works from (added) README.md:
https://github.com/sharetribe/web-template/tree/inquiry-process/src/containers/CheckoutPage

² If you are taking an update from upstream, you might want to check only those translation keys that were added, deleted, or those which got new options for arguments. Those are listed below. The complete list of translation changes to microcopy files looks [like this](https://github.com/sharetribe/web-template/pull/195/files?file-filters%5B%5D=.json&show-deleted-files=true&show-viewed-files=true).

Out of those the most important ones for no-code users are these:

### New translation keys:
```diff
+  "CheckoutPage.default-inquiry.orderBreakdown": "Order breakdown",
+  "CheckoutPage.default-inquiry.title": "Send an inquiry to {authorDisplayName}",

+  "CheckoutPageWithInquiryProcess.messageLabel": "Inquiry message",
+  "CheckoutPageWithInquiryProcess.messagePlaceholder": "Hello there! I'm interested in…",
+  "CheckoutPageWithInquiryProcess.messageRequired": "Message is required.",
[...]

PR #196, opened 2023-08-04
Fix: crash on checkout page (after inquiry)

EugeneLoy:
When rendering `CheckoutPage`, the variable `tx`, (that is used to calculate breakdown and pay in) is preferred to be based on `existingTransaction`, rather than `speculatedTransaction`.

As a result, when trying to checkout after inquiry (as a customer, (1) visit listing page, (2) send message to provider and (3) request booking from order page) `CheckoutPage` crashes on render with the following output in console:

Capture

This is because `existingTransaction` does not yet have all data expected by `getFormattedTotalPrice`.

This PR makes pay in/breakdown calculation prefer speculated transaction, plus some minor cosmetic refactor to make related code clearer.

### DIFF
--- a/src/containers/CheckoutPage/CheckoutPage.js
+++ b/src/containers/CheckoutPage/CheckoutPage.js
@@ -762,9 +762,7 @@ export class CheckoutPageComponent extends Component {
       return <NamedRedirect name="ListingPage" params={params} />;
     }
 
-    // Show breakdown only when (speculated?) transaction is loaded
-    // (i.e. it has an id and lineItems)
-    const tx = existingTransaction.id ? existingTransaction : speculatedTransaction;
+    const tx = speculatedTransaction.id ? speculatedTransaction : existingTransaction;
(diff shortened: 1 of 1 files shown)


Gnito:
Uh, this is embarrassing. Thank you for making PR, it gave clear clue what's happening.

I ended up creating another PR, because the CheckoutPage should prefer actual transaction over speculated one. So, if existing transaction has line-items, it has gone through one of the request-payment transitions - and it's therefore more valid than speculated transaction (which is there to test if the real transaction can be transitioned with payment information).

Note: there can also be problems that come from those 5 sequential XHR calls that CheckoutPage does after submitting the form. So, the real transaction might already have actual line-items in place.

I'm closing this for now.

P.S. we are adding a new process to the Template and we ended up heavily refactoring CheckoutPage because of that.
I added the changes also to this new branch.
https://github.com/sharetribe/web-template/pull/195

PR #216, opened 2023-09-11
Fix: Add namespace to listing field inputs to avoid conflicts in input names

Gnito:
Add namespace prefix to form inputs that represent listing fields to avoid conflicts in input names. (e.g. "pub_" or "priv_")

This fixes the error that happens if a listing field is named "description" - as the EditListingDetailsForm already contains an input named description.

Note: Currently, this only happens if someone adds a listing field with a name _title_, or _description_.


shareoc (approved):
LGTM

PR #223, opened 2023-09-15
Handle missing hosted conf to show maintenance mode page

SariSaar:
This change fixes the current situation where, in case of missing hosted configurations, Maintenance Mode page doesn't get shown because these two functions throw an error.


Gnito (approved):
LGTM!

PR #226, opened 2023-09-20
Added support for all HTTP Methods (GET, POST, PUT, DELETE)

akshatsoni8055:


Gnito:
Thank you for the pull request. It makes sense to improve the support for other HTTP methods, but I felt that it's better to 
1) support the parameter signature that window.fetch has and
2) include the post method (as that might unnecessarily break other customizers setup, when they take update from upstream).

Work-in-progress PR:
https://github.com/sharetribe/web-template/pull/229

PR #235, opened 2023-10-02
Fix: macOS Sonoma seems to have issues with time zone handling

Gnito:
Chrome and Firefox seem to have issues on macOS Sonoma (v14), when populating timeZone on Intl API
- https://bugs.chromium.org/p/chromium/issues/detail?id=1487920
- https://bugzilla.mozilla.org/show_bug.cgi?id=1856428

The bigger problem remaining is the question of whether the Date object is created correctly if the time zone is wrong or undefined.


src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityExceptionForm/ExceptionDateTimeRange.js
```diff
@@ -353,7 +353,13 @@ const ExceptionDateTimeRange = props => {
   const endDateDisabled = !exceptionStartDate || !exceptionStartTime;
   const endTimeDisabled = !exceptionStartDate || !exceptionStartTime || !exceptionEndDate;
   const nextBoundary = findNextBoundary(TODAY, 'hour', timeZone);
-  const placeholderTime = formatDateIntoPartials(nextBoundary, intl, { timeZone })?.time;
+  const placeholderTime = '08:00';
+  try {
+    placeholderTime = formatDateIntoPartials(nextBoundary, intl, { timeZone })?.time;
```
SariSaar, line 358: Can't assign to `const`

PR #245, opened 2023-10-16
Fix wrong pending state name at checkout page

ugened47:
`PAYMENT_PENDING` is a no-existing state so `shouldFetchPaymentIntent` is always `false` and `onRetrievePaymentIntent` is never called.


Gnito:
@ugened47  Uh, nice catch - and thank you for reporting!

I created another PR and included this commit there. 
I'm closing this as I'm merging the 246.

PR #253, opened 2023-10-19
Fix pagination limit

Gnito:
The gist of this pull request is this:

```js
  const pageCountLimit = paginationUnsupported
    ? 1
    : hasPaginationLimit
    ? paginationLimit
    : totalPages;
```


bladealslayer (approved):
LGTM!

PR #259, opened 2023-10-26
Use the correct translation asset in email templates

SariSaar:


Gnito:
@SariSaar Add changelog entry. Otherwise, I don't really have anything to add.

PR #271, opened 2023-11-06
final-form: add extend polyfill through a patch.

Gnito:
Some ppl experience this with their setup, since iOS **17.0.x** is unreliably setting Object.assign() support.

Note: **this should be fixed on iOS _17.1_**

This essentially follows this comment:
https://github.com/final-form/final-form/issues/186#issuecomment-1779109218


pkonduri:
This fixes my longstanding issue on iPhone12 + iOS 17 where the app crashes when navigating from listing to search back to listing

PR #291, opened 2024-01-11
So users only see line items relevant to their role

jayenashar:


Gnito:
Thanks for the PR and sorry that I haven't responded earlier. 

This needed a bit of product vision discussion: a long time ago (when the component was created), we were not sure if some role-specific _line-items_ should be also shown to the other party. E.g. if a marketplace offers insurance as an additional item/fee for customers, then it might make sense that the provider sees it too. Anyway, we've now come to the conclusion that it does not make sense to show that kind of information through receipts => so, filtering makes sense.

Gnito:
I refactored the code a bit + added a changelog entry. Those are now in a separate PR - so, I'm closing this one.

PR #293, opened 2024-01-15
Customercommission

SariSaar:


server/api-util/lineItems.js
```diff
@@ -194,9 +194,25 @@ exports.transactionLineItems = (listing, orderData, providerCommission) => {
...
+        {
+          code: 'line-item/customer-commission',
+          unitPrice: calculateTotalFromLineItems([order]),
+          percentage: customerCommission.percentage,
+          includeFor: ['customer'],
+        },
+      ]
+    : [];
+
   // Let's keep the base price (order) as first line item and provider's commission as last one.
```
Gnito, line 208: Update code comment

server/api-util/lineItemHelpers.js
```diff
@@ -268,7 +268,7 @@ exports.constructValidLineItems = lineItems => {
  * @returns boolean
  */
 exports.hasCommissionPercentage = commission => {
-  const percentage = commission.percentage;
+  const percentage = commission?.percentage;
```
Gnito, line 271: 👍

server/api-util/lineItems.js
```diff
@@ -182,7 +182,7 @@ exports.transactionLineItems = (listing, orderData, providerCommission) => {
   };
 
   // Note: extraLineItems for product selling (aka shipping fee)
-  //       is not included to commission calculation.
+  //       is not included in either customer or provider commission calculation.
```
Gnito, line 185: Perhaps there should be a bit longer description of what provider vs customer commissions are - and how they affect the payin and payout.

server/api-util/lineItemHelpers.js
```diff
@@ -268,11 +268,14 @@ exports.constructValidLineItems = lineItems => {
...
   const isDefined = percentage != null;
   const isNumber = typeof percentage === 'number' && !isNaN(percentage);
   if (isDefined && !isNumber) {
     throw new Error(`${percentage} is not a number.`);
   }
-  return isDefined;
+
+  // Only create a line item if the percentage is not zero
+  const isZero = percentage === 0;
+  return isDefined && !isZero;
```
Gnito, line 280: It's a number at this point, so perhaps we should use 'greater than'. At least for me, it's more understandable that the commission should be more than zero (instead of not-zero).

```suggestion
  // Only create a line item if the percentage is set to be more than zero
  const isMoreThanZero = percentage > 0;
  return isDefined && isMoreThanZero;
```

SariSaar: Fair point, updated!

Gnito (approved):
LGTM

PR #318, opened 2024-02-12
So parameter customerCommission is described in JSDoc

jayenashar:


Gnito (approved):
Looks good to me!

Thanks for the PR.
Side note: usually, we require a changelog entry to be added (as the first entry line in CHANGELOG list). This is so small that I didn't add it this time.

PR #326, opened 2024-02-21
SDK upgrade + other updates

lyyder:
In this PR:

- Upgrade SDK to 1.20.1
- Improved docstrings and code comments in Facebook and Google auth code
- Fix showing IdP auth errors in the login page:
Screenshot 2024-02-21 at 14 02 39


server/api/auth/google.js
```diff
@@ -78,9 +99,22 @@ exports.authenticateGoogle = (req, res, next) => {
   })(req, res, next);
 };
 
-// Use custom callback for calling loginWithIdp enpoint
-// to log in the user to Sharetribe marketplace with the data from Google
+/** This function is called when user returns to this application after authenticating with
```
Gnito, line 102:
```suggestion
/**
 * This function is called when user returns to this application after authenticating with
```

server/api/auth/facebook.js
```diff
@@ -27,6 +27,21 @@ const strategyOptions = {
...
 };
 
+/**
+ * Function Passport calls when a redirect returns the user from Facebook to the application.
+ *
+ * Normally with Passport, this function is used to validate the received user data, and possibly
+ * create a new user and the `done` callback is a session management function provided by Passport.
+ * In our case, the Sharetribe SDK handles user creation and session management. Therefore, we only
+ * extract user data here and provide a custom session management function in
+ * `authenticationFacebookCallback`, that servers as the `done` callback here.
```
Gnito, line 37:
```suggestion
 * `authenticateFacebookCallback `, that servers as the `done` callback here.
```

src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -194,8 +201,12 @@ export const AuthenticationForms = props => {
...
   // eslint-disable-next-line no-confusing-arrow
   const errorMessage = (error, message) => (error ? message : null);
+
+  const loginOrIdpAuthErrorMessage = (loginError, idpAuthError) =>
+    errorMessage(idpAuthError, idpAuthErrorMessage) || errorMessage(loginError, loginErrorMessage);
+
   const loginOrSignupError = isLogin
-    ? errorMessage(loginError, loginErrorMessage)
+    ? loginOrIdpAuthErrorMessage(loginError, idpAuthError)
     : errorMessage(signupError, signupErrorMessage);
```
Gnito, line 210: How about using pure optional chaining instead?

```js
  const hasError = error => !!error;
  const loginOrSignupError = isLogin && hasError(loginError)
    ? loginErrorMessage
    : isLogin && hasError(idpAuthError)
    ? idpAuthErrorMessage
    : hasError(signupError)
    ? signupErrorMessage
    : null;
```

server/api/auth/facebook.js
```diff
@@ -53,25 +68,43 @@ if (clientID) {
...
+ * @param {*} next Call the next middleware function in the stack
+ */
 exports.authenticateFacebookCallback = (req, res, next) => {
+  // We've already defined the `verifyCallback` function for the Passport Facebook authentication
+  // strategy. That function is normally used to verify the user information obtained from identity
+  // provider, or alternatively create a new use while an internal Passport function is used to
+  // store the user data into session. In our case however, we use the SDK to manage sessions.
+  // Therefore, we provide an additional session management function here, that is called from the
+  // `verifyCallback` fn.
   passport.authenticate('facebook', function(err, user) {
```
Gnito, line 108: Did I understand correctly that this anonymous function is the "done" function in verifyCallback?

How about not making it explicit with a function expression?
```js
const doneForVerifyCallback = function(err, user) {
  loginWithIdp(err, user, req, res, clientID, 'facebook');
};
passport.authenticate('facebook', doneForVerifyCallback)(req, res, next);
```

lyyder: If I'm correct, yes. Or it may be that there are some functions in between the `done` from `verifyCallback` and this. There might be Passports own user serialization fns. The Passport.js documentation is not exactly explicit on this. Maybe calling this `sessionFn` etc would convey a correct message.

src/containers/AuthenticationPage/AuthenticationPage.js
lyyder: Maybe with the exception that leave out the `hasError` fn and just use `!!` in the chaining.

Gnito (approved):
LGTM

PR #337, opened 2024-03-12
User fields

SariSaar:
This prepares the codebase for user fields. We'll release a new user-fields asset later. 
Note: the official release for this code change is most likely bundled together with the categories feature.

## Add configuration-based custom user field support
- Add custom user fields configuration
  - Supported schema types: enum, multi-enum, text, long, boolean
- Add custom user field validation

## Profile Page
- Show custom user field data from user's extended data on the profile page
- Duplicate custom field components SectionDetailsMaybe, SectionMultiEnumMaybe, and SectionTextMaybe from ListingPage to ProfilePage
- Add new util/fieldHelpers.js file with a general pickCustomFieldProps function that can be used to get props for the Section...Maybe components on ProfilePage and, after updates to listing custom field config, on ListingPage.
- Add tests to ProfilePage.js
- Sanitize configuration based user fields when setting user to state.marketplaceData

## ProfileSettingsPage
- Move CustomExtendedDataField.js to src/components for easier shared use
- Add new util/userHelpers.js file with functions for managing user field data
- Update ProfileSettingsForm.js to show inputs based on configured user fields
- Save custom user field data in the user's extended data from ProfileSettingsPage.js.
- Add tests for ProfileSettingsForm

## AuthenticationPage
- Update SignupForm.js and ConfirmSignupForm.js to show inputs based on configured user fields
- Save custom user field data as part of creating a user on AuthenticationPage.js
- Update SignupForm tests and example
- Add ConfirmSignupForm tests and example

### New translations
```json
  "ProfilePage.detailNo": "No",
  "ProfilePage.detailYes": "Yes",
  "ProfilePage.detailsTitle": "Details",
```


src/containers/ProfilePage/ProfilePage.js
```diff
@@ -142,6 +151,50 @@ export const DesktopReviews = props => {
   );
 };
 
+export const CustomUserFields = props => {
+  const { publicData, metadata, userFields } = props;
+  return (
+    <>
+      <H4 as="h2" className={css.listingsTitle}>
+        <FormattedMessage id="ProfilePage.customFieldsHeading" />
+      </H4>
```
Gnito, line 160: Wireframes don't have this heading included. So, perhaps it's better to omit it.

It's also a bit odd case from the heading structure:

H1: "Hello, I'm Juho M"

  H2: Details
  H2: SectionMultiEnumMaybe / SectionTextMaybe (all of the fields)

If we add an extra heading between H1 and sections, we should add H2 for the customFieldsHeading and H3 for the sections:
H1: "Hello, I'm Juho M"
  H2: customFieldsHeading
    H3: Details
    H3: SectionMultiEnumMaybe / SectionTextMaybe (all of the fields)

--

However, I don't think the extra heading for the custom fields would serve any meaningful purpose since the bio & custom fields could just tell the same story.


H1: "Awesome Pasta Joint"
p: "Our restaurant is located in central Helsinki and serves fresh pasta. We make pasta dough on the morning. [etc...]"

H2 (text): Origin story
p: "We came up with the idea of creating own restaurant during the pandemic [etc...]"

H2 (multi-enum): Pasta types we use
- Spaghetti
- Penne
- Riggatone
- Etc...

Gnito: So, I'd suggest removing this heading (and then check that the margins work without it).

src/config/configUser.js
```diff
@@ -0,0 +1,121 @@
+export const userFields = [
```
Gnito, line 1: If this configUser is included to codebase, add also a similar code-comment as listingFields have. 
I.e. introduction to the structure of the config and what are the possible values that the components can handle.

src/config/configUser.js
```diff
@@ -0,0 +1,121 @@
...
+    userTypeConfig: {
+      limitToUserTypeIds: true,
+      userTypeIds: ['a', 'b', 'c'],
+    },
+  },
+  {
+    key: 'canCook',
+    label: 'Can you cook?',
+    scope: 'public',
+    schemaType: 'boolean',
```
Gnito, line 28: Am I correct that we have everything available in the codebase for supporting **boolean** and **long** schema types for user fields?

We haven't added those schematypes for the listing fields, because we have not created filter components for them yet. However, user fields don't need filter components.

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -142,6 +151,50 @@ export const DesktopReviews = props => {
...
+                key={key}
+                heading={config?.label}
+                options={createFilterOptions(enumOptions)}
+                selectedOptions={value || []}
+              />,
+            ]
+          : isTargetListingType && hasValue && config.schemaType === SCHEMA_TYPE_TEXT
+          ? [...pickedElements, <SectionTextMaybe key={key} heading={config?.label} text={value} />]
+          : pickedElements;
+      }, [])}
```
Gnito, line 193: Perhaps we could extract the props generation into a new function (added to some util file)

```js
const pickCustomFieldProps = (publicData, metadata, entityTypeKey, activeEntityTypes=[]) => {
  return userFields.reduce((pickedElementProps, config) => {
        const { key, enumOptions, includeForListingTypes, scope = 'public', schemaType } = config;
        // TODO fix isTargetListingType => isTargetUserType
        const entityType = publicData?.[entityTypeKey];
        const isTargetEntityType =
          activeEntityTypes == null || activeEntityTypes.includes(entityType);

        const createFilterOptions = options =>
          options.map(o => ({ key: `${o.option}`, label: o.label }));

        const value =
          scope === 'public' && !!publicData
            ? publicData[key]
            : scope === 'metadata' && !!metadata
            ? metadata[key]
            : null;

        const hasValue = value != null;
        return isTargetEntityType && schemaType === SCHEMA_TYPE_MULTI_ENUM
          ? [
              ...pickedElementProps,
              {
                schemaType,
                key,
                heading: config?.label,
                options: createFilterOptions(enumOptions),
                selectedOptions: value || [],
              },
            ]
          : isTargetEntityType && hasValue && config.schemaType === SCHEMA_TYPE_TEXT
          ? [...pickedElementProps, { schemaType, key heading: config?.label, text: value }]
          : pickedElementProps;
      }, []);
```

Then we could use that here and on ListingPage(s)

```js
// before JSX return statement: 
// const propsForCustomFields = pickCustomFieldProps(publicData, metadata, 'userType', includeForUserTypes);
// or
// const propsForCustomFields = pickCustomFieldProps(publicData, metadata, 'listingType', includeForListingTypes);

// Inside JSX, we could have more straightforward code:
{propsForCustomFields.map(customFieldProps => {
  const { schemaType, ...fieldProps } = customFieldProps;
  return schemaType === SCHEMA_TYPE_MULTI_ENUM
    ? <SectionMultiEnumMaybe {...fieldProps} />
    : schemaType === SCHEMA_TYPE_TEXT
    ? <SectionMultiEnumMaybe {...fieldProps} />
    : null;
})}
```

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -142,6 +151,50 @@ export const DesktopReviews = props => {
...
+      <H4 as="h2" className={css.listingsTitle}>
+        <FormattedMessage id="ProfilePage.customFieldsHeading" />
+      </H4>
+      <SectionDetailsMaybe {...props} />
+      {userFields.reduce((pickedElements, config) => {
+        const { key, enumOptions, includeForListingTypes, scope = 'public' } = config;
+        // TODO fix isTargetListingType => isTargetUserType
+        const listingType = publicData?.listingType;
+        const isTargetListingType =
+          includeForListingTypes == null || includeForListingTypes.includes(listingType);
```
Gnito, line 167: At minimum, I think this should be handled (or commented out with exposed `const isTargetListingType = true;`)
However, check my larger comment on this setup.

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -231,7 +295,16 @@ const ProfilePageComponent = props => {
...
         footer={<FooterContainer />}
       >
-        <MainContent bio={bio} displayName={displayName} userShowError={userShowError} {...rest} />
+        <MainContent
+          bio={bio}
+          displayName={displayName}
+          userShowError={userShowError}
+          publicData={publicData}
+          metadata={metadata}
+          userFields={userFields}
```
Gnito, line 304: Perhaps this should be explicitly named as config? E.g. "userFieldConfigs"

src/util/configHelpers.js
```diff
@@ -671,6 +787,53 @@ const restructureListingFields = hostedListingFields => {
...
+      return key
+        ? {
+            key,
+            scope,
+            schemaType,
+            ...enumOptionsMaybe,
+            filterConfig: {
+              ...filterConfig,
+              label: filterConfig.label || defaultLabel,
+            },
```
Gnito, line 820: There's no filterConfig for user fields. We should not add it (and make sure the code elsewhere doesn't expect this.

src/util/configHelpers.js
```diff
@@ -722,6 +885,18 @@ const mergeListingConfig = (hostedConfig, defaultConfigs) => {
   };
 };
 
+const mergeUserConfig = (hostedConfig, defaultConfigs) => {
+  const hostedUserFields = restructureUserFields(hostedConfig?.userFields?.userFields);
+
+  const { userFields: defaultUserFields } = defaultConfigs.user;
+
+  // TODO: add shouldMerge logic
```
Gnito, line 893: Prepare this first part (config & show on profile page) as feature complete. I.e. avoid leaving TODOs behind.

src/containers/ProfilePage/ProfilePage.test.js
```diff
@@ -0,0 +1,100 @@
...
+  createCurrentUser,
+  createListing,
+  createReview,
+  createUser,
+  fakeIntl,
+  fakeViewport,
+} from '../../util/testData';
+import { renderWithProviders as render, testingLibrary } from '../../util/testHelpers';
+
+import { ProfilePageComponent } from './ProfilePage';
```
Gnito, line 14: This is OKis, but did you consider trying to test the _actual_ ProfilePage. The benefit would be that also configs (and the relevant part of configHelper.js) would be tested at the same time.

src/containers/ProfilePage/ProfilePage.module.css
```diff
@@ -203,3 +203,105 @@
...
+    margin-bottom: 40px;
+  }
+}
+
+/* 
+ * Custom field component classes 
+ */
+
+.sectionHeading,
+.sectionHeadingWithExtraMargin {
```
Gnito, line 222: is this `.sectionHeadingWithExtraMargin` used in somewhere in these components?

src/containers/ProfilePage/ProfilePage.module.css
```diff
@@ -203,3 +203,105 @@
...
+  /* Preserve newlines, but collapse other whitespace */
+  white-space: pre-line;
+
+  @media (--viewportMedium) {
+    padding: 3px 0 5px 0;
+    margin: 0;
+  }
+}
+
+.ingress {
```
Gnito, line 257: I don't this is used in these components

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -142,6 +151,50 @@ export const DesktopReviews = props => {
   );
 };
 
+export const CustomUserFields = props => {
+  const { publicData, metadata, userFields } = props;
+  return (
+    <>
+      <H4 as="h2" className={css.listingsTitle}>
```
Gnito, line 158: This doesn't sound right (but I'd remove the whole heading as argued in a different comment).

src/util/configHelpers.js
```diff
@@ -539,6 +601,60 @@ const validListingFields = (listingFields, listingTypesInUse) => {
...
+          console.warn(`Unsupported user extended data configurations detected (${name}) in`, data);
+        }
+
+        return { config: { ...acc.config, ...prop }, isValid: hasFoundValid };
+      },
+      { config: {}, isValid: true }
+    );
+
+    if (validationData.isValid) {
+      const hasIncludeForListingTypes = validationData.config?.includeForListingTypes;
```
Gnito, line 646: This should be about user types not listing types.

---

P.S:
The idea has been that if config does not include any limitation to smaller set of (listing|user) types, then we automatically fill in all the active (user|listing) types. 
=> the end result is that the code can rely on the existence of includeFor(Listing|User)Types.

P.P.S
I've been considering if we should change this behavior so that there's a known value for it (like "ALL")
I.e. both versions would be allowed:
- `includeForListingTypes: ['a', 'b'],`
- `includeForListingTypes: 'all',`

Basically, because categories would also need this kind of behavior and there can be 
a) gazillion categories and 
b) categories can be nested (a flat array doesn't make that much sense with them).

src/util/configHelpers.js
```diff
@@ -305,6 +305,23 @@ const validKey = (key, allKeys) => {
   return [isUniqueKey, { key }];
 };
 
+const validUserTypesForUserConfig = (includeForUserTypes, userTypesInUse) => {
```
Gnito, line 308: You have this function, but you don't use it anywhere.

src/translations/en.json
```diff
@@ -610,7 +610,11 @@
   "ProductOrderForm.selectDeliveryMethodOption": "Select delivery method",
   "ProductOrderForm.selectQuantityOption": "Select...",
   "ProductOrderForm.shippingOption": "Shipping",
+  "ProfilePage.customFieldsHeading": "User information",
```
Gnito, line 613: If heading is removed, remember to remove this too.

src/config/configUser.js
SariSaar: Yes, both SectionDetailsMaybe and CustomExtendedDataField support boolean and long already.

Gnito: I pinged Juho & Thoma about this possibility (but I don't expect that we would reveal long&boolean through the Console in a different pace with listing fields - it probably makes a better release story to have them both work the same way)

src/containers/ProfilePage/ProfilePage.test.js
SariSaar: Is there an example page where the test targets the actual page, so I can check what the difference is?

src/containers/ProfilePage/ProfilePage.module.css
SariSaar: SectionTextMaybe has the option of showing the text block as ingress with this class, so if we want to remove that capability from the ProfilePage version, then we can also remove this class.

src/containers/ProfilePage/ProfilePage.test.js
Gnito: @SariSaar ListingCard is probably the easiest to read 
(but it uses snapshots - don't use snapshots on page component level)
[ListingCard.test.js](https://github.com/sharetribe/web-template/blob/main/src/components/ListingCard/ListingCard.test.js)

Another example could be EditListingPage.test.js:
- [Create config using util/testData.js](https://github.com/sharetribe/web-template/blob/main/src/containers/EditListingPage/EditListingPage.test.js#L153-L164)
- And then [render with created config](https://github.com/sharetribe/web-template/blob/main/src/containers/EditListingPage/EditListingPage.test.js#L261)

src/util/configHelpers.js
SariSaar: Would this then mean that
- the default user conf would still have this structure, and the restructured data would have a top-level includeForUserTypes array
```
userTypeConfig: {
      limitToUserTypeIds: true,
      userTypeIds: ['a', 'b', 'c'],
    },
```
- or is limitToUserTypeIds: true/false an obsolete attribute?

src/containers/ProfilePage/ProfilePage.js
SariSaar: This is a good idea, however it would require either the listing config and user config having the same name for the "includeForListing|UserType" attribute, or then adding some `(if entitykey == listingType then includeForListingType else includeForUserType)` logic. I think we need to finalise this once we land on some config from Console, i.e. after Mikko's vacation.

src/config/configUser.js
SariSaar: Ok. Would it still make sense to have that capability on the template side, even if Console does not yet support it? It could be helpful for customizers.

src/containers/ProfilePage/ProfilePage.js
SariSaar: Heading removed!

src/containers/ProfilePage/ProfilePage.js
SariSaar: I moved this section from ProfilePage to util/fieldHelpers.js now, but I didn't yet start working on combining user and listing capabilities i.e. adding it to listing page.

src/containers/ProfilePage/ProfilePage.js
SariSaar: Addressed in moving this to helper!

src/util/configHelpers.js
SariSaar: Roger that, removed!

src/containers/ProfilePage/ProfilePage.js
SariSaar: Heading removed!

src/translations/en.json
SariSaar: Removed!

src/config/configUser.js
Gnito: Yes, definitely!
This was more about making a note that after this PR is merged, it only needs Console changes to take boolean and long schema types into use for user fields.

src/containers/ProfilePage/ProfilePage.js
Gnito: @SariSaar Uh, good point! I've been too deep in the filters & EditListingWizard code, that I'm not anymore remembering how the asset data is currently structured.

Mikko might have some opinion about the grouped checkboxes (showConfig&saveConfig), but this structure should not change, as any change there would be inconsistent with the existing listingTypeConfig: 

```
(user|listing)TypeConfig: {
      limitTo(User|listing)TypeIds: true,
      userTypeIds: ['a', 'b', 'c'],
}
```

...and now that I need to add a similar limiter for categories too. It's probably best that I also remove **includeForListingType** property in favor of listingTypeConfig.

src/util/configHelpers.js
Gnito: @SariSaar Good point (as I said in the other thread). Forget what I said about includeForListingTypes, I'll swap it to a more verbose **listingTypeConfig** wrapper.

src/containers/ProfilePage/ProfilePage.test.js
SariSaar: The test suite has now been updated to target ProfilePage instead of ProfilePageComponent!

src/util/configHelpers.js
SariSaar: Updated user field validation to use this! If no user types are passed to this function, it sets the valid value to 

```
      userTypeConfig: {
        limitToUserTypeIds: false,
      },
```

src/util/configHelpers.js
SariSaar: This has now been updated!

src/containers/ProfilePage/ProfilePage.js
SariSaar: The helper has now been generalized and should be able to handle both user type and listing type restrictions

src/containers/ProfilePage/SectionDetailsMaybe.js
```diff
@@ -0,0 +1,63 @@
...
+const SectionDetailsMaybe = props => {
+  const { publicData, metadata, userFieldConfig, intl } = props;
+
+  if (!publicData || !userFieldConfig) {
+    return null;
+  }
+
+  const pickUserFields = (filteredConfigs, config) => {
+    const { key, schemaType, enumOptions, userConfig = {}, showConfig = {} } = config;
+    const { limitToUserTypeIds, userTypeIds } = userConfig;
```
Gnito, line 16: This sounds like a typo? Isn't this supposed to be _userTypeConfig_?

SariSaar: Welp, good catch! Updated now.

src/containers/ProfileSettingsPage/ProfileSettingsForm/ProfileSettingsForm.js
```diff
@@ -19,9 +20,11 @@ import {
   IconSpinner,
   FieldTextInput,
   H4,
+  CustomExtendedDataField,
 } from '../../../components';
 
 import css from './ProfileSettingsForm.module.css';
+import { getPropsForCustomUserFieldInputs } from '../../../util/userHelpers';
```
Gnito, line 27: @SariSaar move this to among other **util** imports (to avoid circular references).

Gnito: Note: since userHelpers import types, this import should be after that import statement:
`import { propTypes } from '../../../util/types';`

jayenashar:
@IshanVeer and i just added saveConfig.defaultValue to this.  please consider adding it before release.

and a side question: why do you force-push?

Gnito:
@jayenashar interesting concept, but at least at this point of testing, we don't add more features. I'll add this as a feature request, though.

That being said, there's a flip side to this coin.  Adding more input options complicates the UI, and we want to avoid making the Console side too overwhelming for no-code ppl.

---

We use _rebase workflow_ to 
- keep the commit history of the main branch clean and linear. This way
  - debugging is easier
  - it's also easier to inspect history and how code has evolved (which is good for us as well as customizers who work the repository and take updates from there).
- it also makes it possible to cleanly incorporate new code into a feature branch

jayenashar:
> I'll add this as a feature request, though.

Thanks.

>       * it's also easier to inspect history and how code has evolved (which is good for us as well as customizers who work the repository and take updates from there).

I find that rebasing changes history and therefore isn't a true reflection of how code has evolved.  E.g., if I find a bug and try to trace it back to a commit, with rebasing, I often trace it to the wrong commit because that commit has moved it's place in the tree and the bug wasn't there when the commit was written and is actually caused by a merge.

>     * it also makes it possible to cleanly incorporate new code into a feature branch

I'm not sure I understood this point.  If you rebase a feature branch onto new code, don't you need to re-resolve conflicts?  The tree is cleaner for sure, but the process seems messier to me.

Sorry for derailing this PR!

Gnito:
> I find that rebasing changes history and therefore isn't a true reflection of how code has evolved.

It is true for the main branch and the main branch is never rebased. Devs can do whatever they want on feature branches, but what is merged to the main, is the final record - and here only the main branch is considered relevant for the production or customizers. 

If bugs end up in the main branch (which is a natural part of development), then it's clear to pick one of the past merge-commits to be deployed. The reason, why it's clear, is that the merge-commit of the feature has a clear scope: the dev, who made it, has not merged other features branches or bug fixes into it and conflicts are resolved early on instead of when the merge happens. So, the process is to rebase the feature branch on top of the main branch, when the main receives those other (tested and approved) features and bug-fixes. Essentially, the rebase workflow enforces an individual developer to build, test, and work out code differences at the feature branch level before merging to the main branch.

Rebase workflow is used quite often in bigger teams that deploy big codebases to production - library projects do that more rarely. ...but let's not deep-dive into rebase workflow here - as there are several articles about pros and cons. 
- https://www.atlassian.com/git/tutorials/merging-vs-rebasing
- https://dev.to/smitterhane/git-things-right-using-rebase-workflow-in-a-software-team-jjp
- http://kensheedlo.com/essays/why-you-should-use-a-rebase-workflow/
- https://git-scm.com/book/en/v2/Git-Branching-Rebasing#_rebase_vs_merge

PR #338, opened 2024-03-12
Dailight saving time: fix Atlantic/Azores tz issue with getStartOf fn call

Gnito:
EditListingAvailbilityExceptionForm: the call for **getStartOf** function with unit 'day' creates an infinite loop.

The underlying issue is that the Moment library returns a bad time as the start of the day:
https://github.com/moment/moment-timezone/issues/409

If we look for the start of the day on _'Atlantic/Azores'_ time zone, Moment returns 23:00 hours (March 30) because of DST (since 00:00 does not exist for Azores). There are a couple of other time zones that work like this too.

Most time zones use 03:00 to avoid this issue with DST.
The fix (for the infinite loop) is to add the 1-day offset as before **+ X hours more** and then ask for _startOf_ 'day' (or week/month). After that, the _startOf_ then returns 01:00 (March 31) aka the actual start of the day.

### DIFF
--- a/src/util/dates.js
+++ b/src/util/dates.js
@@ -265,7 +265,31 @@ export const getStartOf = (date, unit, timeZone, offset = 0, offsetUnit = 'days'
     : moment(date).clone();
 
   const startOfUnit = m.startOf(unit);
-  const startOfUnitWithOffset = offset === 0 ? startOfUnit : startOfUnit.add(offset, offsetUnit);
+  // Note: there is an issue with the Moment library when dealing with
+  //  Daylight Saving Times (DST) in the 'Atlantic/Azores' time zone.
+  //  When calculating the start of the day that follows March 30th, Moment
+  // returns 23:00 on the same day due to the switch in DST. The point
+  // in time 00:00 does not exist in this timezone when DST takes effect.
+  // This creates an infinite loop in the code as it expects to receive the
+  // date following the date that was queried. A couple of other time
+  // zones apply DST around midnight too. Most time zones apply DST
+  // at 03:00 to avoid this issue.
+  //
+  // The fix (for the infinite loop) is to ask for the next day and then add extra hours
+  // (10 hours) to that date. After that, we can ask the start of the day using
+  // startOf 'day' (or week/month). Using this logic, calculating the
+  // start of the next day in the Atlantic/Azores timezone on the 30th of March,
+  // this returns 01:00 (March 31st), which is the actual start of the next day.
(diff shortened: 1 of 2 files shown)


src/util/dates.js
```diff
@@ -265,7 +265,23 @@ export const getStartOf = (date, unit, timeZone, offset = 0, offsetUnit = 'days'
...
   const startOfUnit = m.startOf(unit);
-  const startOfUnitWithOffset = offset === 0 ? startOfUnit : startOfUnit.add(offset, offsetUnit);
+  // Note: if we look for the start of the day on _'Atlantic/Azores'_ time zone,
+  // Moment returns 23:00 hours (March 30) because of DST (since 00:00 does not exist for Azores).
+  // There are a couple of other time zones that work like this too.
+  // Most time zones use 03:00 to avoid this issue with DST.
+  // The fix (for the infinite loop) is to add the 1-day offset as before + X hours more
+  // and then ask for _startOf_ 'day' (or week/month). After that, the _startOf_ call
+  // then returns 01:00 (March 31) aka the actual start of the day.
+  // https://github.com/moment/moment-timezone/issues/409
```
shareoc, line 275:
```suggestion
// Note: there is an issue with the Moment library when dealing with
//  Daylight Saving Times (DST) in the 'Atlantic/Azores' time zone.
//  When calculating the start of the day that follows March 30th, Moment 
// returns 23:00 on the same day due to the switch in DST. The point
// in time 00:00 does not exist in this timezone when DST takes effect.
// This creates an infinite loop in the code as it expects to receive the 
// date following the date that was queried. A couple of other time 
// zones apply DST around midnight too. Most time zones apply DST
// at 03:00 to avoid this issue.
//
// The fix (for the infinite loop) is to add a 1 day offset by adding 
// 10 hours to the date and then calculating the start of the day using
//  _startOf_ 'day' (or week/month). Using this logic, calculating the 
// start of the next day in the Atlantic/Azores timezone on the 30th of March,
// this correctly returns 01:00 (March 31st), which is the actual start of the next day.
// For more information, see: https://github.com/moment/moment-timezone/issues/409
```

Gnito: I'll change the latter part:

// The fix (for the infinite loop) is to ask for the next day and then add extra hours 
// (10 hours) to that date. After that, we can ask the start of the day using
//  _startOf_ 'day' (or week/month). Using this logic, calculating the 
// start of the next day in the Atlantic/Azores timezone on the 30th of March,
// this returns 01:00 (March 31st), which is the actual start of the next day.

PR #342, opened 2024-03-19
So the graph matches the .edn

jayenashar:


Gnito:
There has been a unification process to change several words to match the US word forms, but the default-booking process still uses _**state/cancelled**_ vocabulary.

https://github.com/sharetribe/web-template/blob/main/ext/transaction-processes/default-booking/process.edn#L116

jayenashar:
Yes, but the js file doesn't match

Gnito:
@jayenashar ah, right! I'm blind or something because I read the change wrongly :D

Perhaps you could change the changelog entry.
`The state in the default-booking/process.edn is spelled the American way.`
=> `The state in the default-booking/process.edn is still spelled the British English way with 2 "l's".`

jayenashar:
Sorry, I did have that backwards!  Being an American living in Australia, I no longer know which English is from England!

Although now I think this PR should be closed.  Changing the name in the graph js means we also need to change the marketplace texts.  We could change the edn, but a) then it would be out of sync with the default process when you create a new marketplace; and b) it doesn't matter because the edn state name is not used anywhere.

PR #346, opened 2024-03-21
Save user fields on AuthenticationPage

SariSaar:


src/containers/AuthenticationPage/AuthenticationPage.js
```diff
@@ -489,6 +547,7 @@ export const AuthenticationPageComponent = props => {
               idpAuthError={authError}
               signupError={signupError}
               confirmError={confirmError}
+              userFields={userFields}
```
Gnito, line 550: This is passed quite deeply here. Perhaps it would make more sense to get the config directly where it's needed:
`const config = useConfiguration();`

src/util/userHelpers.js
```diff
@@ -85,32 +85,44 @@ export const initialValuesForUserFields = (data, targetScope, targetUserType, us
...
+ * is a signup form. Defaults to true.
+ * @returns an array of props for CustomExtendedDataField: key, name,
+ * fieldConfig, defaultRequiredMessage
  */
 
-export const getPropsForCustomUserFieldInputs = (userFieldsConfig, intl, userType = null) => {
-  return userFieldsConfig.reduce((pickedFields, fieldConfig) => {
-    const { key, userTypeConfig, schemaType, scope } = fieldConfig || {};
-    const namespacedKey = getNamespacedKey(scope, key);
+export const getPropsForCustomUserFieldInputs = (
```
Gnito, line 96: Remove the empty line between this and the JSDoc-comment

src/containers/AuthenticationPage/AuthenticationPage.js
SariSaar: Roger, set useConfiguration() to the parent components of each form!

PR #351, opened 2024-03-28
Fixes to email templates

shareoc:
Some email templates were not using the format-text helper which supports localization


SariSaar (approved):
LGTM!

PR #364, opened 2024-04-19
Add long filter

shareoc:


src/containers/SearchPage/SearchPageWithMap.js
```diff
@@ -321,6 +321,24 @@ export class SearchPageComponent extends Component {
       ...customSecondaryFilters,
     ];
 
+    // Inject this temporary long filter for testing
```
Gnito, line 324: Inject through configListing:
- https://github.com/sharetribe/web-template/blob/main/src/config/configListing.js#L49
- https://github.com/sharetribe/web-template/blob/main/src/util/configHelpers.js#L904

Then check that the min & max are valid
https://github.com/sharetribe/web-template/blob/main/src/util/configHelpers.js#L607
Something like this:
```js
            : name === 'min'
            ? validMin(value, schemaType)
            : name === 'max'
            ? validMax(value, schemaType, data)
```

They should be 
- Integers ([Number.isInteger()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger))
- Between [MIN_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER) and [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER), and
- Max should be bigger than min (which is why I added data aka config to validMax in the example)

And the validation check should be added for listing fields as well as user fields

src/containers/SearchPage/FilterComponent.js
```diff
@@ -1,15 +1,12 @@
 import React from 'react';
-
-import { SCHEMA_TYPE_ENUM, SCHEMA_TYPE_MULTI_ENUM } from '../../util/types';
-import { constructQueryParamName } from './SearchPage.shared';
+import { SCHEMA_TYPE_ENUM, SCHEMA_TYPE_MULTI_ENUM, SCHEMA_TYPE_LONG } from '../../util/types';
```
Gnito, line 2: Actually, don't remove the empty lines. Instead, add comments around different imports.
Something like this:
https://github.com/sharetribe/web-template/blob/e16b014ae6461df71704e421c53f14e506869b1b/src/index.js#L36

(This was initiated after we started to receive issues that customizers didn't pay attention to imports and started to receive errors.)

So, we should probably give a bit more emphasis to the order of imports. I.e.  if imports are made in wrong bad order they might cause circular dependency issues. Therefore, it would be a good practice to add comments like `// utils`,  `// components` while touching old components.

Here's more about this problem scope:
https://github.com/sharetribe/web-template/blob/e16b014ae6461df71704e421c53f14e506869b1b/src/components/index.js#L5-L10

src/components/CustomExtendedDataField/CustomExtendedDataField.js
```diff
@@ -94,15 +94,32 @@ const CustomFieldText = props => {
...
+const validateNumber = (value, max, min, numberTooSmallMessage, numberTooBigMessage) => {
+  const parsedValue = Number.parseInt(value, 10);
+  if (parsedValue > max) {
+    return numberTooBigMessage;
+  }
+  if (parsedValue < min) {
+    return numberTooSmallMessage; 
+  }
+  return undefined; 
+};
```
Gnito, line 106: Move this to util/validators.js

Gnito: So, that it can be used as 

`validate={validators.validateInterger(max, min, numberTooSmallMessage, numberTooBigMessage)}`

Note: validateNumber is too loaded name. Number in JavaScript is double precision floating point number. We specifically need to check if the input is integer.

src/components/CustomExtendedDataField/CustomExtendedDataField.js
```diff
@@ -117,7 +134,7 @@ const CustomFieldLong = props => {
       }}
       label={label}
       placeholder={placeholder}
-      {...validateMaybe}
+      validate={value => validate(value, maximum, minimum)}
```
Gnito, line 137: Check my other comment.
`validate={validators.validateInterger(max, min, numberTooSmallMessage, numberTooBigMessage)}`

Gnito: perhaps the order: `minimum, maximum` is more clean.

src/components/CustomExtendedDataField/CustomExtendedDataField.js
```diff
@@ -94,15 +94,32 @@ const CustomFieldText = props => {
...
-  const validateMaybe = isRequired
-    ? { validate: required(requiredMessage || defaultRequiredMessage) }
-    : {};
   const placeholder =
     placeholderMessage || intl.formatMessage({ id: 'CustomExtendedDataField.placeholderLong' });
+  const numberTooSmallMessage = intl.formatMessage({ id: 'CustomExtendedDataField.numberTooSmall'}, { min: minimum });
+  const numberTooBigMessage = intl.formatMessage({ id: 'CustomExtendedDataField.numberTooBig'}, {max: maximum} );
+
+    const validate = (value, max, min) => {
+      const requiredError = isRequired && !value ? (requiredMessage || defaultRequiredMessage) : undefined;
```
Gnito, line 119: If value is 0 this fails.
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

Use `value != null` as that checks if value is something else than **null** or **undefined**.

src/containers/SearchPage/SelectNumberFromRangeFilter/SelectNumberFromRangeFilter.module.css
```diff
@@ -0,0 +1,100 @@
...
+  padding-bottom: 5px;
+  margin-top: 0px;
+}
+@media (--viewportMedium) {
+  .contentWrapper {
+    padding-top: 0px;
+    padding-bottom: 8px;
+    margin-top: 0px;
+  }
+}
```
Gnito, line 31: Nest this media query inside .contentWrapper (and not the other way around).

src/containers/SearchPage/SelectNumberFromRangeFilter/SelectNumberFromRangeFilter.module.css
```diff
@@ -0,0 +1,100 @@
...
+  @media (--viewportMedium) {
+    margin: 5px 8px 0 8px;
+  }
+}
+
+.sliderWrapper {
+  display: flex;
+  padding: 17px 0 6px 0;
+}
+@media (--viewportMedium) {
```
Gnito, line 96: Move this media query inside the sliderWrapper.

src/util/configHelpers.js
```diff
@@ -1049,7 +1077,7 @@ const mergeListingConfig = (hostedConfig, defaultConfigs, categoriesInUse) => {
 
   // When debugging, include default configs by passing 'true' here.
   // Otherwise, use listing types and fields from hosted assets.
-  const shouldMerge = mergeDefaultTypesAndFieldsForDebugging(false);
+  const shouldMerge = mergeDefaultTypesAndFieldsForDebugging(true);
```
Gnito, line 1080: Don't commit this one

src/containers/SearchPage/SelectNumberFromRangeFilter/SelectNumberFromRangeFilter.js
```diff
@@ -0,0 +1,169 @@
...
+      initialValues={resolvedInitialValues}
+      {...rest}
+    >
+      <Field
+        max={max}
+        min={min}
+        name={name}
+        step={step}
+        component={RangeInput}
+      />
```
Gnito, line 119: Perhaps it's best that Field and RangeInput are on the same file. 

Then this would be:
`<FieldSelectIntegerRange name={name} min={min} max={max} step={step} />`

Inside the file, you could have:

```jsx
const RangeInput = props => {
 // ...
}
const FieldSelectIntegerRange = props => {
  return (
    <Field
      max={max}
      min={min}
      name={name}
      step={step}
      component={RangeInput}
    />
  );
}
export default FieldSelectIntegerRange;
```

So, I suggest that you rename this component as **FieldSelectIntegerRange**

src/containers/SearchPage/SelectNumberFromRangeFilter/SelectNumberFromRangeFilter.js
```diff
@@ -0,0 +1,169 @@
...
+// Transforms a range object containing `minValue` and `maxValue`
+// into query parameter format, i.e. "0,1000"
+const formatToQueryParam = (range, queryParamName) => {
+  const { minValue, maxValue } = range || {};
+  // We explicitly compare `minValue` to null as comparing to '0' returns false but '0' is a valid range value.
+  const value = minValue != null && maxValue != null ? `${minValue},${maxValue}` : null;
+  return { [queryParamName]: value };
+};
+
+const SelectNumberFromRangeFilter = props => {
```
Gnito, line 37: I'd suggest that you rename this component and file.

This does not select a number from a range. It selects a range and the range needs to be an integer. 
(Btw. number in JavaScript is a double by default ([double precision floating point](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_encoding)).

Suggestion: **IntegerRangeFilter**

src/components/CustomExtendedDataField/CustomExtendedDataField.js
shareoc: Should the validate function also be in the utils file? I'm referring to this one

```
   const validate = (value, min, max) => {
      const requiredError = isRequired && value != null ? undefined : (requiredMessage || defaultRequiredMessage);
      const numberError = validateInteger(value, max, min, numberTooSmallMessage, numberTooBigMessage);
      return requiredError || numberError; 
    };
```

 I feel like it could also be left here as this specific validation logic is only relevant to this single component.

src/util/configHelpers.js
shareoc: Hmm, should I amend the original commit?

src/components/CustomExtendedDataField/CustomExtendedDataField.js
Gnito: The util/validators.js is the default place where customizers should try to find validators for React Final Form Fields.

Other validation needs can be more ad hoc.

src/util/configHelpers.js
Gnito: Yes. Create a new fixup commit and merge it using interactive rebasing. (I can help)

Gnito:
I still try to find time to test the debounce issue.

shareoc:
@Gnito Ok, the debounce issue is solved

src/containers/SearchPage/IntegerRangeFilter/IntegerRangeFilter.module.css
```diff
@@ -0,0 +1,97 @@
...
+    margin: 5px 8px 0 8px;
+  }
+}
+
+.sliderWrapper {
+  display: flex;
+  padding: 17px 0 6px 0;
+  @media (--viewportMedium) {
+    padding: 3px 0 6px 0;
+  }
```
Gnito, line 96:
```suggestion
  padding: 17px 0 25px 0;
  @media (--viewportMedium) {
    padding: 3px 0 13px 0;
  }
```

The distance between "clear" and slider is too short.

This way the element gets better height:
- 66px height on the mobile layout (divisible by baseline of 6px)
- 40px height on the desktop layout (divisible by baseline of 8px)

Especially, with thumb-based navigation on mobile devices would risk accidentally clicking the 'clear' button.

Gnito:
@shareoc Could you fix the bugs that I have created. I.e. with the current setup, you can't pass 0 as a value for the 'long' type field:

https://github.com/sharetribe/web-template/blob/add-long-filter/src/containers/EditListingPage/EditListingWizard/EditListingDetailsPanel/EditListingDetailsPanel.js#L116
```js
const fieldValue = getFieldValue(data, namespacedKey);
```

https://github.com/sharetribe/web-template/blob/add-long-filter/src/containers/EditListingPage/EditListingWizard/EditListingDetailsPanel/EditListingDetailsPanel.js#L169
```js
const fieldValue = getFieldValue(data, key);
```

And import getFieldValue from util/fieldHelpers.js

src/containers/SearchPage/IntegerRangeFilter/FieldSelectIntegerRange.js
```diff
@@ -0,0 +1,141 @@
...
+
+import css from './IntegerRangeFilter.module.css';
+
+const RADIX = 10;
+
+//  Ensures that the given values object has both minimum and maximum values.
+//  If `values` does not include `minValue` or `maxValue`, defaults are applied.
+const resolveMinMaxValues = (values, defaultMax, defaultMin) => {
+  const { maxValue, minValue } = values || {};
+  return { maxValue: maxValue || defaultMax, minValue: minValue || defaultMin };
```
Gnito, line 17: This causes an issue when minValue or maxValue is 0.

```suggestion
  return { 
    maxValue: maxValue != null ? maxValue : defaultMax, 
    minValue: minValue != null ? minValue : defaultMin,
  };
```

Gnito:
I still think there's some update issues with the inputs. Especially, when working with negative values.

You might have been correct that those inputs should have their own states and they should perhaps only update the field-level's state, when the input value is a valid integer.

PR #387, opened 2024-05-16
Handle listings with draft and pending-approval state with login-as

Gnito:


walterrw (approved):
Looks good to me, except I'm not sure about decoding the query param. So the query params are not already decoded at that point, the query parsing doesn't decode them?

PR #398, opened 2024-05-27
Update SDK to v1.21.1 - Fix the issue with extended data with `length` key and a number value

rap1ds:
Update SDK to v1.21.1 - Fix the issue with extended data with `length` key and a number value


Gnito (approved):
LGTM

PR #406, opened 2024-06-06
Remove extra backslash from email template

SariSaar:
The extra backslash causes the translation helper to not work properly, and results in the template printing the translation helper instead of the message.

Before:
Screenshot 2024-06-06 at 11 28 28

After:
Screenshot 2024-06-06 at 11 28 41


Gnito (approved):
LGTM

PR #440, opened 2024-08-23
update the order of the configuration script

shareoc:
Updated the wording in the Mapbox configuration script from 

> Mapbox is the default map provider of the application. Sign up for Mapbox and go the to account page. Then click Create access token. For more information see the: Integrating to map providers documentation.
> If you don't set the Mapbox key, the map components won't work in the application.

to this:

> The Mapbox access token should be set via the Sharetribe Console. You can also configure it as an environment variable, which will be used as a fallback value.

Also moved the prompt on `REACT_APP_MARKETPLACE_NAME` to the mandatory variables section.

### DIFF
--- a/scripts/config.js
+++ b/scripts/config.js
@@ -114,6 +114,10 @@ const mandatoryVariables = settings => {
     settings && settings.REACT_APP_MAPBOX_ACCESS_TOKEN !== ''
       ? { default: settings.REACT_APP_MAPBOX_ACCESS_TOKEN }
       : {};
+  const marketplaceNameDefaultMaybe =
+    settings && settings.REACT_APP_MARKETPLACE_NAME !== ''
+      ? { default: settings.REACT_APP_MARKETPLACE_NAME }
+      : {};
 
   return [
     {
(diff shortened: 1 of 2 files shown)


scripts/config.js
```diff
@@ -140,6 +136,21 @@ ${chalk.dim(
   `Client Secret is needed for privileged transitions with Marketplace API.
 Client secret is connected to client ID. You can find your client secret from Sharetribe Console.`
 )}
+`,
+      default: function() {
+        return rootUrlDefault ? rootUrlDefault : 'http://localhost:3000';
```
Gnito, line 141: I think you have copy-pasted the defaults wrong between 
- SHARETRIBE_SDK_CLIENT_SECRET
- REACT_APP_MARKETPLACE_NAME
- REACT_APP_MARKETPLACE_ROOT_URL

scripts/config.js
```diff
@@ -205,26 +209,21 @@ ${chalk.dim(
 )}
 `,
       default: function() {
-        return rootUrlDefault ? rootUrlDefault : 'http://localhost:3000';
+        return 'MyMarketplace';
```
Gnito, line 212: Don't change this.
The end result is `REACT_APP_MARKETPLACE_ROOT_URL=MyMarketplace
`, which is wrong.

Gnito (approved):
LGTM!

PR #444, opened 2024-09-02
Make Block's propType blockId optional

OtterleyW:
This PR makes the propType `blockId` optional for all Block types. This follows the changes in Console where the Block ID is no longer a required field.


Gnito (approved):
LGTM!

PR #450, opened 2024-09-11
Add transaction initiate rights handling

SariSaar:
This PR prepares the codebase for the ability to restrict transaction rights per individual user. Operators can allow or restrict users from initiating transactions, independent of other actions that the user can take on the marketplace.

When the user does not have this permission in their **effectivePermissionSet** relationship and they try to make an order or send an inquiry, they are redirected to the NoAccessPage.

Screenshot 2024-09-11 at 10 25 12

### New translations:

One of the new translations uses rich text formatting for the FormatMessage – [read more here](https://formatjs.io/docs/react-intl/upgrade-guide-3x/#enhanced-formattedmessage--formatmessage-rich-text-formatting).

```
  "InquiryForm.noTransactionRightsError": "Oops, something went wrong. You don't have transaction rights. <NoAccessLink>Read more about transaction rights.</NoAccessLink>",
  "NoAccessPage.initiateTransactions.schemaTitle": "No transaction rights",
  "NoAccessPage.initiateTransactions.heading": "You don't have transaction rights",
  "NoAccessPage.initiateTransactions.content": "To start a transaction, you need to receive transaction rights from the {marketplaceName} team.",
```

## Notes

For developers who see this before the feature is fully released in the API & Docs:

When **_transaction rights have been revoked_** for a marketplace user, the following Marketplace API endpoint will be restricted:

**Transactions**
POST "transactions/initiate"


src/util/userHelpers.js
```diff
@@ -167,3 +167,20 @@ export const isUserAuthorized = (currentUser, permissionsToCheck) => {
...
     : isActive;
 };
+
+/**
+ * Check if currentUser has permission to post listings.
+ *
+ * @param {Object} currentUser API entity
+ * @returns {Boolean} true if currentUser has permission to post listings.
+ */
+export const hasPermissionToInitiateTransactions = currentUser => {
```
SariSaar, line 177: This logic assumes that once the Core side is done, all users get `initiateTransactions: 'permission/allow'` in their effectivePermissionSet unless explicitly prohibited in the UI. 

Is that a correct assumption, or should this check the accessControl.json asset to see whether the marketplace restricts initiating transactions or not?

src/util/urlHelpers.js
```diff
@@ -19,6 +19,7 @@ export const PROFILE_PAGE_PENDING_APPROVAL_VARIANT = 'pending-approval';
 
 // No access page - path params:
 export const NO_ACCESS_PAGE_POST_LISTINGS = 'posting-right';
+export const NO_ACCESS_PAGE_INITIATE_TRANSACTIONS = 'transaction-right';
```
Gnito, line 22: So, the URL would be /no-transaction-right
This is a user facing string, so good to think if there are something more understandable.
(But personally, I couldn't think anything better...)

src/translations/en.json
```diff
@@ -511,6 +511,9 @@
   "NoAccessPage.postListings.schemaTitle": "No publishing rights",
   "NoAccessPage.postListings.heading": "You can't publish listings",
   "NoAccessPage.postListings.content": "You need to receive publishing rights from the {marketplaceName} team.",
+  "NoAccessPage.initiateTransactions.schemaTitle": "No transaction rights",
+  "NoAccessPage.initiateTransactions.heading": "You can't initiate transactions",
+  "NoAccessPage.initiateTransactions.content": "You need to receive transaction rights from the {marketplaceName} team.",
```
Gnito, line 516: Btw. were these approved by Katri (I think she has lately been the one proofreading the translations)

Gnito:
@SariSaar you should make a check to CheckoutPage too.
I.e user might be on a CheckoutPage, when operator removes the right to initiate transactions.
https://github.com/sharetribe/web-template/pull/428/
- https://github.com/sharetribe/web-template/pull/428/commits/035c124294fe6fe5d123d0aed00cb8bbb3c9749f

Also consider other scenarios:
A user is already on contact form (on listing page) and operator removes the right before the user submits the form
https://github.com/sharetribe/web-template/pull/428/commits/788292c24f9728677444c07695889211f65fb006

Related to this, I don't think you should add the redirection to the contact forms submit aka **handleSubmitInquiry**.
If the redirection is there, user will lose the what they were writing to that form.
Better UX is that
a) We don't allow user to open the modal on handleContactUser function. https://github.com/sharetribe/web-template/pull/428/commits/c967d74fc78874be73b2e578bd39c07f88f948eb
b) if the form is open (while operator removes the right), we handle the error that API returns. https://github.com/sharetribe/web-template/pull/428/commits/788292c24f9728677444c07695889211f65fb006

This way user is proactively redirected to NoAccessPage and in the rare scenario where restriction is added while user is writing a long message, they don't lose that message.

src/containers/ListingPage/ListingPage.shared.js
```diff
@@ -138,21 +142,37 @@ export const handleContactUser = parameters => () => {
...
-  const { message } = values;
+  const {
+    history,
+    params,
+    getListing,
+    onSendInquiry,
+    routes,
+    setInquiryModalOpen,
+    currentUser,
+  } = parameters;
```
Gnito, line 153: Check my other comments. My suggestion is that you don't make changes to this function.

src/containers/ListingPage/ListingPage.shared.js
```diff
@@ -173,66 +193,71 @@ export const handleSubmit = parameters => values => {
...
-    bookingDates,
-    bookingStartTime,
-    bookingEndTime,
-    bookingStartDate, // not relevant (omit)
-    bookingEndDate, // not relevant (omit)
-    quantity: quantityRaw,
-    deliveryMethod,
-    ...otherOrderData
-  } = values;
+  if (!hasPermissionToInitiateTransactions(currentUser)) {
```
Gnito, line 196: Perhaps we should not add changes to this function either: as handleSubmit does not remove the need to check the rights on CheckoutPage. If we do it there, does this become irrelevant?
Something like this is already on CheckoutPage: https://github.com/sharetribe/web-template/commit/035c124294fe6fe5d123d0aed00cb8bbb3c9749f

src/translations/en.json
SariSaar: These texts were from the most recent wireframes, but I'll double check!

src/util/urlHelpers.js
SariSaar: I asked Katri's opinion on this in the same wireframe comment where I'm double checking the copy for the no-access page.

src/containers/ListingPage/ListingPage.shared.js
SariSaar: Updated and moved the check to CheckoutPage!

src/containers/ListingPage/ListingPage.shared.js
SariSaar: Updated and moved the check to the handleContactUser function!

src/util/userHelpers.js
SariSaar: Function moved!

As far as the permission check, I don't have a super strong opinion. My intuition is that when we have a larger array of these different permissions, then customizers may want to create different kinds of behaviors for different use cases, in which case it would be useful for them if each function only does a single thing and not a big isUserAuthorized function that wraps all that behavior out of sight. It might also help folks understand what exactly is being allowed and restricted when they read the code.

Having atomic functinos would also guide the devs into making explicit decisions on what permissions go in what features and not rely on the isUserAuthorized function to make the decisions for them. 

If we decide to wrap things into the isUserAuthorized function, then folks can of course modify the logic there, too, so either way works for me.

src/containers/ListingPage/ListingPage.shared.js
SariSaar: The way this now works on CheckoutPage if the operator restricts the user mid-initiating transaction is that the user sees the error message, and when they refresh, they are redirected. Should the checkout page redirect already as a response to the 403 error from the API?

src/util/urlHelpers.js
SariSaar: URLs updated to use plural for both posting and transaction rights, i.e.
- /no-posting-rights
- /no-transaction-rights

src/util/userHelpers.js
SariSaar: After an IRL conversation we ended up simplifying the isUserAuthorized helper function to just check currentUser state and not accept permissionsToCheck param, so those changes have been implemented!

src/containers/NoAccessPage/NoAccessPage.test.js
```diff
@@ -11,10 +11,10 @@ const { screen, userEvent } = testingLibrary;
 const noop = () => null;
 
 describe('NoAccessPageComponent', () => {
-  it('Check that /no-posting-right has heading and content', () => {
+  it('Check that /no-posting-rights has heading and content', () => {
```
Gnito, line 14: Perhaps, you could add another test for the /no-transaction-rights

SariSaar: Test added!

src/translations/en.json
```diff
@@ -511,6 +512,9 @@
   "NoAccessPage.postListings.schemaTitle": "No publishing rights",
   "NoAccessPage.postListings.heading": "You can't publish listings",
   "NoAccessPage.postListings.content": "You need to receive publishing rights from the {marketplaceName} team.",
+  "NoAccessPage.initiateTransactions.schemaTitle": "No transaction rights",
+  "NoAccessPage.initiateTransactions.heading": "You don’t have transaction rights",
+  "NoAccessPage.initiateTransactions.content": "You need to receive transaction rights from the {marketplaceName} team.",
```
Gnito, line 517: Enforce alphabetical order.

Gnito (approved):
LGTM
(but reorder en.json)

src/translations/en.json
```diff
@@ -511,6 +512,9 @@
   "NoAccessPage.postListings.schemaTitle": "No publishing rights",
   "NoAccessPage.postListings.heading": "You can't publish listings",
   "NoAccessPage.postListings.content": "You need to receive publishing rights from the {marketplaceName} team.",
+  "NoAccessPage.initiateTransactions.schemaTitle": "No transaction rights",
+  "NoAccessPage.initiateTransactions.heading": "You don’t have transaction rights",
```
thomasmalbaux, line 516: The apostrophe should be `'` (instead of `’`), so that it is consistent with other places we use this symbol.

src/translations/en.json
thomasmalbaux, line 517: I'm confused with the back-and-forth regarding the copy texts, but here I believe that it is not using Katri's final version at https://whimsical.com/permissions-access-control-LRn1pRNPJ65xhReX5pgcAs@2bsEvpTYSt1His9DMeWRBED1VDEdofji3fK which, at the moment of writing, is `To start a transaction, you need to receive transaction rights from the {marketplaceName} team.` (the first part of the sentence is missing).

thomasmalbaux: Or maybe this is something that will get updated in the CTA project, but I'm not certain of that, so maybe better to have the final text added now already?

src/translations/en.json
```diff
@@ -396,6 +396,7 @@
   "InquiryForm.messageLabel": "Message to {authorDisplayName}",
   "InquiryForm.messagePlaceholder": "Hello {authorDisplayName}! I'm looking forward to…",
   "InquiryForm.messageRequired": "A message is required",
+  "InquiryForm.noTransactionRightsError": "Oops, something went wrong. You don't have permission to make inquiries. <NoAccessLink>Read more about transaction rights.</NoAccessLink>",
```
thomasmalbaux, line 399: I'm not sure if Katri was asked to review this text and I find it a bit odd (even if accurate) that it says "inquiries". I would just use the same "You don't have transaction rights" here, to use fewer concepts and not force end-users to have to understand what is an inquiry and what is a transaction.

Gnito: If/When this is changed, then I think the related message should also changed:
`"InquiryForm.userPendingApprovalError": "Oops, something went wrong. You don't have permission to make inquiries",
`
Although, that message is only shown, if we or customizers create a bug that reveals the modal (or customizers embed the form directly to the listing page without a modal).

src/translations/en.json
SariSaar: Fixed!

src/translations/en.json
SariSaar: Fixed!

src/translations/en.json
SariSaar: I've now changed these two marketplace texts to 

```
"InquiryForm.noTransactionRightsError": "Oops, something went wrong. You don't have transaction rights. <NoAccessLink>Read more about transaction rights.</NoAccessLink>",
"InquiryForm.userPendingApprovalError": "Oops, something went wrong. Your account is waiting for approval.",
```

PR #459, opened 2024-09-30
Viewing rights

SariSaar:
This PR prepares the template for handling viewing rights, denoted by the **read** permission in the user's **effectivePermissionSet** relationship.

When a user's viewing rights have been revoked:
- SearchPage redirects to NoAccessPage
- ListingPage redirects to NoAccessPage unless the user is viewing their own listing. Reviews are not displayed.
- ProfilePage redirects to NoAccessPage unless the user is viewing their own profile. Reviews are not displayed.
- TransactionPage does not show the OrderPanel for transactions at the inquiry state of booking or purchase processes.

Screenshot 2024-10-01 at 12 21 06

## New translations

```
  "NoAccessPage.viewListings.schemaTitle": "No viewing rights",
  "NoAccessPage.viewListings.heading": "You don't have viewing rights",
  "NoAccessPage.viewListings.content": "To view this content, you need to receive viewing rights from the {marketplaceName} team.",
```

## Notes

For developers who see this before the feature is fully released in the API & Docs:

When viewing rights have been revoked for a marketplace user, the following Marketplace API endpoints will be restricted:

**Listings**
GET listings/query
GET listings/show

**Reviews**
GET reviews/query
GET reviews/show

**Timeslots**
GET timeslots/query


src/components/ListingCard/ListingCard.js
```diff
@@ -152,7 +152,7 @@ ListingCardComponent.propTypes = {
   className: string,
   rootClassName: string,
   intl: intlShape.isRequired,
-  listing: propTypes.listing.isRequired,
+  listing: oneOfType([propTypes.listing, propTypes.ownListing]).isRequired,
```
SariSaar, line 155: ListingCard is used on ProfilePage, and if we want to fetch own listings to show on that page, we need to use the ownListings endpoint which yields ownListings resources.

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -252,7 +254,7 @@ export const MainContent = props => {
           </ul>
         </div>
       ) : null}
-      {isMobileLayout ? (
+      {hideReviews ? null : isMobileLayout ? (
```
SariSaar, line 257: I'm assuming here that if the user doesn't have viewing rights i.e. can't see reviews, we want to hide the whole review component instead of showing the review headings with 0 reviews. Let me know if that assumption is correct, or if we do want to show the review headings here either way!

Screenshot 2024-10-01 at 12 13 27

src/util/urlHelpers.js
```diff
@@ -22,6 +22,8 @@ export const NO_ACCESS_PAGE_POST_LISTINGS = 'posting-rights';
 export const NO_ACCESS_PAGE_INITIATE_TRANSACTIONS = 'transaction-rights';
 // If user account is on pending-approval state, then user can't initiate transactions or create listings
 export const NO_ACCESS_PAGE_USER_PENDING_APPROVAL = 'user-approval';
+// If user account has no viewing rights on a private marketplace, they can only see their own listings
+export const NO_ACCESS_PAGE_VIEW_LISTINGS = 'viewing-right';
```
Gnito, line 26: no-viewing-right vs no-viewing-rights?

src/util/userHelpers.js
```diff
@@ -176,3 +176,18 @@ export const hasPermissionToInitiateTransactions = currentUser => {
...
+ * @returns {Boolean} true if currentUser has permission to view listing and user data on a private marketplace.
+ */
+export const hasReadPermission = currentUser => {
+  if (currentUser?.id && !currentUser?.effectivePermissionSet?.id) {
+    console.warn(
+      '"effectivePermissionSet" relationship is not defined or included to the fetched currentUser entity.'
+    );
+  }
+  return currentUser?.effectivePermissionSet?.attributes?.read === 'permission/allow';
+};
```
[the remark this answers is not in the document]
Gnito: It might also make sense to add a link to API reference > currentUser's PermissionSet 
https://www.sharetribe.com/api-reference/marketplace.html#currentuser-permissionset

...but I'm not sure when it will get every permission mentioned there.

src/containers/ListingPage/ListingPage.duck.js
```diff
@@ -406,10 +410,14 @@ export const loadData = (params, search, config) => (dispatch, getState, sdk) =>
...
-    dispatch(fetchReviews(listingId)),
-  ]).then(response => {
+  const hasViewingRights = currentUser && hasReadPermission(currentUser);
+  const promises = hasViewingRights
+    ? // For users with viewing rights, fetch the listing and the associated reviews
+      [dispatch(showListing(listingId, config)), dispatch(fetchReviews(listingId))]
+    : // If user has no viewing rights, only allow fetching their own listing without reviews
+      [dispatch(showListing(listingId, config, true))];
+
+  return Promise.all(promises).then(response => {
```
Gnito, line 420: You need to restrict also fetchMonthlyTimeSlots call that happens in the "then" scope.

src/containers/ListingPage/ListingPageCarousel.js
```diff
@@ -496,17 +499,25 @@ const EnhancedListingPage = props => {
...
   const hasUserPendingApprovalError = isErrorUserPendingApproval(showListingError);
+
   if ((isPrivateMarketplace && isUnauthorizedUser) || hasUserPendingApprovalError) {
     return (
       <NamedRedirect
         name="NoAccessPage"
         params={{ missingAccessRight: NO_ACCESS_PAGE_USER_PENDING_APPROVAL }}
       />
     );
+  } else if (hasNoViewingRights && isForbiddenError(showListingError)) {
```
Gnito, line 514: My assumption is that this should be || not &&

src/containers/ListingPage/ListingPageCoverPhoto.js
```diff
@@ -494,17 +497,25 @@ const EnhancedListingPage = props => {
...
   const hasUserPendingApprovalError = isErrorUserPendingApproval(showListingError);
+
   if ((isPrivateMarketplace && isUnauthorizedUser) || hasUserPendingApprovalError) {
     return (
       <NamedRedirect
         name="NoAccessPage"
         params={{ missingAccessRight: NO_ACCESS_PAGE_USER_PENDING_APPROVAL }}
       />
     );
+  } else if (hasNoViewingRights && isForbiddenError(showListingError)) {
```
Gnito, line 512: Same assumption as on ListingPageCarousel (`||`)

src/containers/ListingPage/ListingPageCarousel.js
Gnito: You could also add a check for "isPrivateMarketplace &&" as above.

The point is that we want to make sure that on what conditions these redirections happen. Devs (future us & customizers) should not prevent SSR on listing pages on public marketplaces.

src/containers/NoAccessPage/NoAccessPage.js
```diff
@@ -58,6 +60,12 @@ export const NoAccessPageComponent = props => {
         heading: 'NoAccessPage.initiateTransactions.heading',
         content: 'NoAccessPage.initiateTransactions.content',
       }
+    : isViewingRightsPage
+    ? {
+        schemaTitle: 'NoAccessPage.viewListings.schemaTitle',
+        heading: 'NoAccessPage.viewListings.heading',
+        content: 'NoAccessPage.viewListings.content',
+      }
```
Gnito, line 68: There's probably a mismatch between this and Aleksi's PR.
Perhaps you should rebase your branch on top of it? Although, Aleksi might merge his branch soonish anyway.

src/containers/ProfilePage/ProfilePage.duck.js
```diff
@@ -230,9 +241,20 @@ export const loadData = (params, search, config) => (dispatch, getState, sdk) =>
   // Note 2: In private marketplace mode, this page won't fetch data if the user is unauthorized
   const isAuthorized = currentUser && isUserAuthorized(currentUser);
   const isPrivateMarketplace = config.accessControl.marketplace.private === true;
+  const hasNoViewingRights = currentUser && !hasReadPermission(currentUser);
   const canFetchData = !isPrivateMarketplace || (isPrivateMarketplace && isAuthorized);
+  // On a private marketplace, show current user's own page even if they don't have viewing rights
```
Gnito, line 246: Add 'active' state somehow to this sentence.

"show current user's own page for active aka approved users even if they don't have viewing rights"

src/containers/ProfilePage/ProfilePage.duck.js
```diff
@@ -136,14 +140,21 @@ export const queryUserListings = (userId, config) => (dispatch, getState, sdk) =
...
-    })
+  const queryParams = {
+    include: ['author', 'images'],
+    'fields.image': [`variants.${variantPrefix}`, `variants.${variantPrefix}-2x`],
+    ...createImageVariantConfig(`${variantPrefix}`, 400, aspectRatio),
+    ...createImageVariantConfig(`${variantPrefix}-2x`, 800, aspectRatio),
+  };
+
+  const listingsPromise = ownProfileOnly
+    ? sdk.ownListings.query(queryParams)
```
Gnito, line 151: Add a filter. 

This now includes draft & pending-approval & closed listings too.
Screenshot 2024-10-01 at 13 54 35

src/containers/ProfilePage/ProfilePage.js
Gnito: Makes sense to me 👍

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -294,17 +296,20 @@ export const ProfilePageComponent = props => {
...
   }
 
-  const isDataLoaded = isPreview
-    ? currentUser != null || userShowError != null
-    : user != null || userShowError != null;
   const isCurrentUser = currentUser?.id && currentUser?.id?.uuid === pathParams.id;
+  const isDataLoaded =
+    isPreview || (isCurrentUser || useCurrentUser)
+      ? currentUser != null || userShowError != null
+      : user != null || userShowError != null;
```
Gnito, line 303: This is not understandable.

The original setup tries to say that if page is rendered as a "preview" variant (e.g. user is in pending-approval state), then there must be a currentUser entity (it can't be null or undefined) and no errors when the call was made.

If the site is public, then user entity is the important data.

This now looks very strange and there's a hint of circular reasoning.

How about expanding this ternary to optional chaining setup?

```js
const isDataLoaded = isPreview
  ? currentUser != null || userShowError != null
  : hasNoViewingRightsOnPrivateMarketplace
  ? currentUser != null || userShowError != null
  : user != null || userShowError != null;
```

(and this could be moved below hasNoViewingRightsOnPrivateMarketplace)

src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -427,24 +428,40 @@ export const fetchTransaction = (id, txRole, config) => (dispatch, getState, sdk
...
+        relationships: {
+          ...included[includedListingIdx].relationships,
+          author: {
+            data: {
+              id: provider.id,
+              type: 'user',
+            },
+          },
+        },
+      };
```
Gnito, line 449: Turn this into a helper function like listingRelationship, which is defined on line 377.

src/containers/ListingPage/ListingPageCarousel.js
SariSaar: If this is hasNoViewingRights || isForbiddenError, then it shows the NoAccessPage also when users are trying to reach their own listings.

By default, the .duck.js file is defined to call ownListings endpoint for all listings when a user has restricted viewing rights, and the endpoint throws 403 for everyone else's listings but gives 200 for the current user's own listings. So the combination of noViewingRights && 403 means that the user has tried to view someone else's listings, but they are getting 403 from the ownListings endpoint.

If there's some other way to handle allowing own listings but not other folks' listings, and without fetching the listing first, I can also use that in the .duck.js file.

src/containers/ListingPage/ListingPage.duck.js
SariSaar: Fixed!

src/containers/NoAccessPage/NoAccessPage.js
SariSaar: Rebased on the most recent main and updated to include ctaData

src/util/userHelpers.js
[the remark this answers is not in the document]
SariSaar: Fixed, link added to hasPermission function comments

src/containers/ProfilePage/ProfilePage.duck.js
SariSaar: Comment updated!

src/containers/ProfilePage/ProfilePage.duck.js
SariSaar: Fixed to only include published listings 👍

src/containers/ProfilePage/ProfilePage.js
SariSaar: Updated!

src/containers/TransactionPage/TransactionPage.duck.js
SariSaar: Extracted to a separate helper.

src/containers/ProfilePage/ProfilePage.js
```diff
@@ -328,6 +335,15 @@ export const ProfilePageComponent = props => {
         params={{ missingAccessRight: NO_ACCESS_PAGE_USER_PENDING_APPROVAL }}
       />
     );
+  } else if (!isPreview && hasNoViewingRightsOnPrivateMarketplace && !isCurrentUser) {
```
Gnito, line 338: I started to think that this "if - else if" chain does not work correctly.

```js
  if (!isDataLoaded) {
    // return etc.
  } else if (!isPreview && isNotFoundError(userShowError)) {
    // return etc.
  } else if (!isPreview && (isUnauthorizedOnPrivateMarketplace || hasUserPendingApprovalError)) {
    // return etc.
  } else if (!isPreview && !isCurrentUser && (hasNoViewingRightsOnPrivateMarketplace || isErrorNoViewingRights(userShowError))) {
    // TODO 
    // - Is it possible to create isErrorNoViewingRights error check?
    // - SSR: should only end up here if isErrorNoViewingRights is thrown (hasNoViewingRightsOnPrivateMarketplace contains assumption of currentUser entity)
    // return etc.
  } else if (!isPreview && isForbiddenError(userShowError)) {
    // The generic 403 Forbidden needs to come after more specific 403s aka hasUserPendingApprovalError (and potential isErrorNoViewingRights that doesn't exist yet)
    // return etc.
  } else if (isPreview && mounted && !isCurrentUser) {
    // return etc.
  } else if ((isPreview || isPrivateMarketplace) && !mounted) {
    // return etc.
  }
  
  // This is rendering normal profile page (not preview for pending-approval)
  return (
```

src/containers/ListingPage/ListingPageCarousel.js
Gnito: There is error code: permission-denied-read

SariSaar: If we use the ownListings.show endpoint in ListingPage.duck.js file, the 403 does not come back as permission-denied-read, because the 403 is caused by a non-author attempting ownListings.show, and the error message is `forbidden`. If we use listings.show endpoint in ListingPage.duck.js, it gives 403 permission-denied-read on all listings, including the user's own, so the isOwnListing check cannot be applied after the SDK call.

However, since read rights only apply to private marketplaces, I could set the ListingPage.duck.js hasViewingRights condition to 

`const hasViewingRights = !isPrivateMarketplace || (isPrivateMarketplace && currentUser && hasPermissionToViewData(currentUser));`

and ListingPage[variant].js hasNoViewingRights to

`const hasNoViewingRights = isPrivateMarketplace && currentUser && !hasPermissionToViewData(currentUser);`

This way, the ownListings.show endpoint (in duck.js) and hasNoViewingRights would explicitly only get applied on a private marketplace. That's what happens in practice anyway, as Console doesn't allow revoking read rights on public marketplaces, but it's also clear for customizers.

Would something like this work?

This would address the issues you listed:

- public marketplace: SSR must work without currentUser entity => hasViewingRights check does not apply
  - The aforementioned error: redirection => this won't ever happen
  - Happy case: by-pass this redirection => this always happens
- private marketplace: CSR only => hasViewingRights check applies in duck.js, currentUser already exists
  - own listings: by-pass this redirection => always shown, whether .duck.js file uses listings.show or ownListings.show
  - someone elses listings => viewing rights checked in duck.js
      - has viewing rights: by-pass this redirection => uses listings.show endpoint, no error
      - no viewing rights: redirection => uses ownListings.show endpoint, gets 403

src/containers/ProfilePage/ProfilePage.js
SariSaar: Does SSR apply on a private marketplace?

I tested this approach anyway now to see if it's possible to create isErrorNoViewingRights here. My initial implementation was to not dispatch showUser at all when `canFetchOwnProfileOnly`, but I tested adding the dispatch in the promises array to catch the error and use it in this condition. 

If I add `showUser` dispatch and remove the `hasNoViewingRightsOnPrivateMarketplace` check from `isDataLoaded`, (so loads the page once `user` or `userShowError` exists) the `userShowError` has time to propagate before the page loads, so that I could add an `isErrorNoViewingRights` function. (This means that the isForbiddenError check in the current implementation is unnecessary, because without the showUser dispatch the userShowError always remains null).

However, if SSR doesn't apply on private marketplaces, this would only apply client-side. It would not change the behavior, and instead it would add one extra SDK call that we only make to get the error, because the `canFetchOwnProfileOnly` condition already has different dispatches going on (own listings vs all listings).

So either way works!

src/containers/ListingPage/ListingPageCarousel.js
Gnito: ListingPage & ProfilePage are part of SSR and because the knowledge of private vs public comes from an asset (that can be stale because assets are cached for 5 min in production), we need to consider how SSR behaves on profile, listing & search pages - even if the marketplace is turned to private mode.

If I read the code correctly, the SSR during those stale moments will fetch user on profile page and listing on listing page. I doubt we can avoid it.

> Side note: I think I made a mistake earlier, when I added fetchCurrentUser to the loadData of this page. Luckily, full page loading doesn't fetch currentUser because isAuthenticated flag is false at that point of data loading and the fetch is not done on the server.

SariSaar: Listing pages NoAccessPage condition tested with SSR-setup (i.e. private marketplace set to false on asset and no current user) and set to

`} else if (hasNoViewingRights && isForbiddenError(showListingError)) {`

=> this condition shows NoAccessPage for missing viewing rights – if hasViewingRights resolves to false in .duck.js, the endpoint that gets used is ownListings, and the error from that endpoint is `forbidden` for no current user. Removed the privateMarketplace check so this catches SSR-situations as well.

src/containers/ListingPage/ListingPageCoverPhoto.js
SariSaar: Fixed similarly!

src/containers/ProfilePage/ProfilePage.js
SariSaar: Condition updated to 

`  } else if (
    (!isPreview && hasNoViewingRightsOnPrivateMarketplace && !isCurrentUser) ||
    isErrorNoViewingPermission(userShowError)
  ) {`

and isErrorNoViewingPermission added.

I also added similar handling to search pages:

`  } else if ((isPrivateMarketplace && hasNoViewingRightsUser) || hasNoViewingRightsError) {`

src/containers/ListingPage/ListingPage.duck.js
```diff
@@ -406,15 +410,19 @@ export const loadData = (params, search, config) => (dispatch, getState, sdk) =>
...
-  return Promise.all([
-    dispatch(showListing(listingId, config)),
-    dispatch(fetchReviews(listingId)),
-  ]).then(response => {
+  const hasViewingRights = currentUser && hasPermissionToViewData(currentUser);
+  const promises = hasViewingRights
+    ? // For users with viewing rights, fetch the listing and the associated reviews
+      [dispatch(showListing(listingId, config)), dispatch(fetchReviews(listingId))]
+    : // If user has no viewing rights, only allow fetching their own listing without reviews
+      [dispatch(showListing(listingId, config, true))];
```
Gnito, line 418: Sorry, one more thing.
This can't default to ownListings endpoint on server because then the server can't return correct HTTP headers on public marketplace scenario. I.e. if listing is not found from public endpoint listings.show, the 404 should be set.

SariSaar: Fixed ListingPage.duck.js condition to default to listings.show, and updated the listing pages condition accordingly

src/containers/TransactionPage/TransactionPage.test.js
```diff
@@ -1171,6 +1171,90 @@ describe('TransactionPage', () => {
...
+
+        // Show listing's title only (side card)
+        expect(getAllByText('listing-item title')).toHaveLength(1);
+
+        // Don't show listing pricing
+        expect(screen.queryByText('$55.00')).toBeNull();
+        expect(screen.queryByText('OrderPanel.perUnit')).toBeNull();
+
+        // Don't show BookingTimeForm inputs
+        expect(screen.queryByText('BookingTimeForm.bookingStartTitle')).toBeNull();
```
Gnito, line 1251: Side note: I haven't used `.toBeNull()` expectation. So, another approach is `.not.toBeInTheDocument()`

src/containers/ProfilePage/ProfilePage.test.js
```diff
@@ -161,3 +185,153 @@ describe('ProfilePage', () => {
...
+        setInitialState(),
+        currentUserShowRequest(),
+        showUserRequest(user.id, config),
+        queryListingsRequest(user.id),
+        currentUserShowSuccess(currentUser),
+        addMarketplaceEntities(fakeResponse(user), sanitizeConfig),
+        showUserSuccess(),
+        addMarketplaceEntities(fakeResponse([listing])),
+        queryListingsSuccess(userListingRefs),
+        queryReviewsSuccess(reviews),
```
Gnito, line 242: There are some surprises. (Old code is missing queryReviewsRequest. It might make sense to fix it at some point but perhaps not inside this PR.)

Gnito (approved):
LGTM!

src/containers/TransactionPage/TransactionPage.test.js
SariSaar: I originally tried with `.not.toBeInTheDocument()`, but I kept getting errors that the test engine wasn't able to find the item, i.e. the page was behaving correctly but the test expectation wasn't catching it correctly. On the other hand, the combination of queryByText and toBeNull resolved correctly.

SariSaar: Looks like the issue was with screen.getByText vs screen.queryByText and not the .not.toBeInTheDocument vs .toBeNull. So I could update these to `expect(screen.queryByText(...)).not.toBeInTheDocument()`!

Gnito: @SariSaar I think this is fine. 
It's just more dramatic to be able to ask the big question: "toBe or not.toBe?"

PR #464, opened 2024-10-07
Add support for adding a link to an image

shareoc:


src/containers/PageBuilder/Primitives/Image/Image.js
```diff
@@ -41,17 +42,22 @@ export const FieldImage = React.forwardRef((props, ref) => {
...
-      <ResponsiveImage
-        className={css.fieldImage}
-        ref={ref}
-        alt={alt}
-        image={image}
-        variants={variantNames}
-        sizes={sizes}
-        {...otherProps}
-      />
+      {imageLinkHref ? <Link href={imageLinkHref}>{responsiveImage}</Link> : responsiveImage}
```
Gnito, line 60: Perhaps you could do something like this:
```suggestion
      {imageLinkHref ? <Link href={imageLinkHref} title={alt}>{responsiveImage}</Link> : responsiveImage}
```

When anchor is wrapping an image, the **_alt_** attribute of the image is not selected to be used as "tooltip". Instead, the anchor's content is used, but as that's an img, it won't show anything.

Screenshot 2024-10-07 at 19 00 31

src/containers/PageBuilder/Primitives/Image/Image.js
```diff
@@ -41,17 +42,28 @@ export const FieldImage = React.forwardRef((props, ref) => {
...
-        className={css.fieldImage}
-        ref={ref}
-        alt={alt}
-        image={image}
-        variants={variantNames}
-        sizes={sizes}
-        {...otherProps}
-      />
+      {imageLinkHref ? (
+        <Link href={imageLinkHref} title={alt}>
```
Gnito, line 61: Your IRL question was a good one! I think, it would be beneficial to include another extra parameter

`<Link href={imageLinkHref} title={alt} fieldType={imageLinkFieldType}>`

It's not used now inside the Link component, but it would be pretty annoying to figure out why the info does not go through in these scenarios with nested components.

Side note: you have already extracted **link** from **props** at this point:
`const imageLinkHref = props.link?.href || null;`

Gnito (approved):
LGTM!
(but rebase this on top of main branch before merging)

PR #468, opened 2024-10-11
fix map width bug on window resize

shareoc:
Fixes a small issue where the static map doesn't correctly adapt to window width on resize

### DIFF
--- a/src/containers/ListingPage/ListingPage.module.css
+++ b/src/containers/ListingPage/ListingPage.module.css
@@ -699,6 +699,10 @@
   border: 0;
   cursor: pointer;
 
+  img {
+    width: 100%;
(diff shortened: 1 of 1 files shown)


src/containers/ListingPage/ListingPage.module.css
```diff
@@ -699,6 +699,10 @@
   border: 0;
   cursor: pointer;
 
+  img {
```
Gnito, line 702: This can't use element styling¹ as any `` element (after this code-chunk) is loaded would be affected.

That's why element styling happens in one place: marketplaceDefaults.css.

¹ CSS Modules can't make element-styles unique. It only affects classes.

PR #469, opened 2024-10-14
Align block content

SariSaar:
This PR prepares the template for an alignment option for content page blocks. The alignment option can have one of three values: 
- left
- center
- right

If the block has an alignment set to one of those values, the content of the block is aligned accordingly.

Screenshot 2024-10-14 at 15 53 22


Gnito (approved):
LGTM

PR #474, opened 2024-10-17
Allow discarding a draft listing

SariSaar:
Add support for discarding draft listings on ManageListingsPage. Draft listings now include a link to discard the listing draft:
discard-draft-link

Clicking that link opens a confirmation modal, and confirming the operation in the modal discards the listing.

discard-confirm-modal


src/containers/ManageListingsPage/ManageListingsPage.duck.js
```diff
@@ -143,6 +166,30 @@ const manageListingsPageReducer = (state = initialState, action = {}) => {
       };
     }
 
+    case DISCARD_DRAFT_REQUEST:
+      return {
+        ...state,
+        discardingDraft: payload.listingId,
```
Gnito, line 172: ManageListingPage.js: has something like this:
`actionsInProgressListingId={openingListing || closingListing}`
That probably should handle **discardingDraft** too.

Gnito (approved):
Handle the **actionsInProgressListingId**  but other than that, looks good to me!

src/containers/ManageListingsPage/ManageListingsPage.duck.js
SariSaar: Included!

src/components/IconAlert/IconAlert.js
```diff
@@ -0,0 +1,46 @@
...
+          d="M18.625 36.086c9.643 0 17.46-7.818 17.46-17.461s-7.817-17.46-17.46-17.46-17.46 7.817-17.46 17.46 7.817 17.46 17.46 17.46Z"
+        />
+      </g>
+    </svg>
+  );
+};
+
+IconAlert.defaultProps = { className: null, rootClassName: null };
+
+const { string } = PropTypes;
```
Gnito, line 39: Move this to line 2:

import { string } from 'prop-types';

This is part of very old syntax, which doesn't make sense now. (since code editors know how to extract correct named imports)

src/containers/ManageListingsPage/ManageListingsPage.js
```diff
@@ -173,16 +198,25 @@ export const ManageListingsPageComponent = props => {
...
                 isMenuOpen={!!listingMenuOpen && listingMenuOpen.id.uuid === l.id.uuid}
-                actionsInProgressListingId={openingListing || closingListing}
+                actionsInProgressListingId={openingListing || closingListing || discardingDraft}
                 onToggleMenu={onToggleMenu}
                 onCloseListing={onCloseListing}
                 onOpenListing={handleOpenListing}
+                onDiscardDraft={openDiscardDraftModal}
                 hasOpeningError={openingErrorListingId.uuid === l.id.uuid}
                 hasClosingError={closingErrorListingId.uuid === l.id.uuid}
+                hasDiscardingError={discardingErrorListingId.uuid === l.id.uuid}
```
Gnito, line 208: No need to modify things, but this pattern of has<X>Error starts to be somewhat stretched at this point.

src/containers/ManageListingsPage/ManageListingsPage.duck.js
```diff
@@ -281,6 +344,34 @@ export const openListing = listingId => (dispatch, getState, sdk) => {
...
+      return delay(300).then(() => {
+        Promise.all([
+          dispatch(discardDraftSuccess(discardResponse)),
+          sdk.ownListings.query(params),
+        ]).then(([_, listingResponse]) => {
+          dispatch(addOwnEntities(listingResponse));
+          dispatch(queryListingsSuccess(listingResponse));
+        });
+      });
+    })
```
Gnito, line 369: It seems I understood your question about in-progress wrongly.

I was thinking a bit different flow:
```suggestion
  return sdk.ownListings
    .discardDraft({ id: listingId }, { expand: true })
    .then(discardResponse => {
      // Set delay before removing the draft so that the user
      // notices which listing card gets removed
      return Promise.all([
        delay(300), // ensure 300 ms delay if queryOwnListings is fast
        dispatch(queryOwnListings(queryParams)),
      ]);
    })
    .then(() => {
      dispatch(discardDraftSuccess(discardResponse)),
    })
    .catch(e =>{//...
```

Something like this
1. Call discardDraft
2. Then call promise.all for queryOwnListings with correct params and a delay
  2.1 Let existing queryOwnListings function to handle the query flow updates
  2.2 There seems to be useless reconstruction for the "queryParams" on the current implementation
  2.3 This can rely on [Promise chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises#chaining).
3. Then mark success for the discard action
3.1 No need to call _removeDiscardedDraftListing_

src/containers/ManageListingsPage/ManageListingsPage.duck.js
```diff
@@ -143,6 +166,30 @@ const manageListingsPageReducer = (state = initialState, action = {}) => {
...
 
+    case DISCARD_DRAFT_REQUEST:
+      return {
+        ...state,
+        discardingDraft: payload.listingId,
+        discardingDraftError: null,
+      };
+    case DISCARD_DRAFT_SUCCESS:
+      return {
+        ...removeDiscardedDraftListing(state, payload.data),
```
Gnito, line 177: I don't think this is needed anymore 🤔

src/containers/ManageListingsPage/ManageListingsPage.duck.js
Gnito: If there's need to return the discardResponse (but I don't think there's a need for that):

```js
  let discardRes = null;
  return sdk.ownListings
    .discardDraft({ id: listingId }, { expand: true })
    .then(discardResponse => {
      discardRes = discardResponse;
      // Set delay before removing the draft so that the user
      // notices which listing card gets removed
      return Promise.all([
        delay(300), // ensure 300 ms delay if queryOwnListings is fast
        dispatch(queryOwnListings(queryParams)),
      ]);
    })
    .then(() => {
      dispatch(discardDraftSuccess(discardResponse)),
      return discardResponse;
    })
    .catch(e =>{//...
    ```

File: src/containers/ManageListingsPage/ManageListingsPage.duck.js (hunk 5 above)
Comment (SariSaar) [reply src/containers/ManageListingsPage/ManageListingsPage.duck.js:177]: It still clears the `discardingDraft` id from state even after the removeDiscardedDraftListing function is removed, so I'd keep it here.

File: src/containers/ManageListingsPage/ManageListingsPage.duck.js (hunk 4 above)
Comment (SariSaar) [reply src/containers/ManageListingsPage/ManageListingsPage.duck.js]: Logic fixed!

File: src/components/IconAlert/IconAlert.js (hunk 2 above)
Comment (SariSaar) [reply src/components/IconAlert/IconAlert.js]: Fixed!

File: src/containers/ManageListingsPage/ManageListingsPage.duck.js [hunk 6]
```diff
@@ -281,6 +326,30 @@ export const openListing = listingId => (dispatch, getState, sdk) => {
...
+      // notices which listing gets removed
+      return Promise.all([delay(300), sdk.ownListings.query(queryParams)]);
+    })
+    .then(([_, listingResponse]) => {
+      dispatch(addOwnEntities(listingResponse));
+      dispatch(queryListingsSuccess(listingResponse));
+    })
+    .then(() => {
+      dispatch(discardDraftSuccess());
+    })
```
Gnito, line 347: Merge these then-closures.

SariSaar: Done!

src/containers/ManageListingsPage/ManageListingsPage.duck.js
Gnito: I meant just the `removeDiscardedDraftListing` function.

PR #479, opened 2024-10-22
Add static 404 page for bots that scan (php) vulnerabilities

Gnito:
Filter out bot requests that scan websites for php vulnerabilities etc. This might look in your logs with attempts to find paths like /wp-admin/index.php, //cms/wp-includes/wlwmanifest.xml, etc. 

There's no need to pass those requests to the React app rendering as it causes unnecessary asset fetches to initialize the server-side rendering.

_Note_: you might want to do this on the edge server instead.

404 page for bots that scan php vulnerabilities:
Screenshot 2024-10-22 at 20 04 38

`/.*(\.php|\.php7|\/wp-.*\/.*|cgi-bin.*|htdocs\.rar|htdocs\.zip|root\.7z|root\.rar|root\.zip|www\.7z|www\.rar|wwwroot\.7z)$/`
Screenshot 2024-10-23 at 11 50 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,9 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2024-XX-XX
 
-- [update] Update copy text for Publishing listing permission to Posting listing.
+- [add] Filter out bot requests that scan websites for vulnerabilities before they reach React app.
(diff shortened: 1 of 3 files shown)


bladealslayer (approved):
lgtm! Easy enough to update if we see some other patterns hit hard.

PR #494, opened 2024-11-06
fix: Menu has a regression bug that's visible on ManageListingCard

Gnito:
Menu has a regression bug on UI for mobile layout. It's visible on the ManageListingCard:
Screenshot 2024-11-06 at 13 57 17

The UI should look like this with the 3-dot menu on the ManageListingCard:
Screenshot 2024-11-06 at 13 57 44

The reason why full-width is needed on some occasions:
Screenshot 2024-11-06 at 13 58 10
This is now reached if property _preferScreenWidthOnMobile_ is passed to Menu.

Note: if the 3-dot menu is expanded at some point with more menu items, it's likely that we move towards full-width usage there too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2024-XX-XX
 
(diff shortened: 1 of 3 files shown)


SariSaar (approved):
LGTM!

PR #499, opened 2024-11-11
Add confirmInProgress to authenticationInProgress state check

AleksaButterfly:
## Description
This PR updates the `authenticationInProgress` function in `auth.duck` to include `confirmInProgress` in the state check. This change ensures that if a confirmation process is in progress, it sets the `inProgress` flag in `ConfirmSignupForm`, displaying a spinner to inform the user that the request is ongoing.

## Changes
- Added `confirmInProgress` to the `authenticationInProgress` function to track the confirmation request state along with login, logout, and signup.


Gnito:
Thank you for reporting and making a fix! I cloned this PR and added a changelog entry.
(closing this one)

PR #514, opened 2024-11-29
Seats: filter listings with seats

SariSaar:
Implement filtering listings with seats.

This PR adds new components. There is a new FieldNumberInput component, as well as a new SeatsFilter component that uses FieldNumberInput. The SeatsFilter component has been implemented on both search pages:
- SearchPageWithGrid uses SeatsFilter as a plain component
seats-filter-plain

- SearchPageWithMap uses SeatsFilter as a popup component
seats-filter-popup

This PR also prepares the codebase to use a hosted configuration for seats filtering, once the feature is released in Console.

### New translations

```json
  "FilterComponent.seatsLabel": "Seats",
  "NumberInput.decrementButton": "Decrement the value by one",
  "NumberInput.incrementButton": "Increment the value by one",
  "SeatsFilter.labelSelected":"Seats: {seatCount}",
```


src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -458,3 +466,23 @@ export const createSearchResultSchema = (
...
+  const aWeekFromNow = stringifyDateToISO8601(addTime(today, 7, 'day'));
+  // Get parameters for dates and seats:
+  const datesAndSeatsMaybe =
+    // If both dates and seats are included, pass both
+    dates && seats
+      ? { dates, seats }
+      : // Dates can be queried without seats
+      dates
+      ? { dates }
+      : // Seats cannot be queried without dates – pass a default date range
```
Gnito, line 481: Add the comments above the variable. These inside-the-ternary comments make the optional chaining pretty difficult to read.

src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -458,3 +466,23 @@ export const createSearchResultSchema = (
     },
   };
 };
+
+export const getDatesAndSeatsMaybe = (dates, seats) => {
+  const today = stringifyDateToISO8601(new Date());
+  const aWeekFromNow = stringifyDateToISO8601(addTime(today, 7, 'day'));
```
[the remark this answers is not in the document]
Gnito: Another thing: you need to take the "prolonged" range into use here too. Otherwise the start and end don't cover all the time zones.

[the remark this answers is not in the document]
SariSaar: The prolonged range and UTC are applied to the params after the `datesAndSeatsMaybe` handling, similarly to default date params, which don't have time zone information when they come from the filter. Also, creating new Date() initialises it in UTC by default, so even if I set my browser timezone to Australia, the start date still points to UTC.

Would there still be a use case for defining the dates in explicit UTC and setting prolonged time both here and in the .duck.js?

src/containers/SearchPage/SearchPage.shared.js
SariSaar: Comments moved to a clearer formatting!

src/components/index.js
```diff
@@ -93,6 +93,7 @@ export { default as FieldReviewRating } from './FieldReviewRating/FieldReviewRat
 export { default as FieldSelect } from './FieldSelect/FieldSelect';
 export { default as FieldSelectTree } from './FieldSelectTree/FieldSelectTree';
 export { default as FieldTextInput } from './FieldTextInput/FieldTextInput';
+export { default as FieldNumber } from './FieldNumber/FieldNumber';
```
Gnito, line 96: Suggestion: move this to line 91

src/containers/SearchPage/SearchPageWithGrid.js
```diff
@@ -135,12 +136,17 @@ export class SearchPageComponent extends Component {
         // We should always trust urlQueryParams with those.
         // The same applies to keywords, if the main search type is keyword search.
         const keywordsMaybe = isMainSearchTypeKeywords(config) ? { keywords } : {};
+        const datesAndSeatsMaybe = getDatesAndSeatsMaybe(
+          prevState.currentQueryParams,
```
Gnito, line 140: I'm not entirely sure if this should be `mergedQueryParams` instead of `prevState.currentQueryParams` 
(aka I can't remember if some liveEdit or page load scenario has empty prevState and therefore the "old" values would come entirely from urlQueryParams).

src/containers/SearchPage/SeatsFilter/SeatsFilter.js
```diff
@@ -0,0 +1,162 @@
...
+    );
+  }
+}
+
+SeatsFilter.defaultProps = {
+  rootClassName: null,
+  className: null,
+  initialValues: null,
+  contentPlacementOffset: 0,
+};
```
Gnito, line 143: Let's have a small huddle related to defaultProps.

src/translations/en.json
```diff
@@ -754,6 +755,7 @@
   "SearchPage.schemaDescription": "Showing search results",
   "SearchPage.schemaForSearch": "a query",
   "SearchPage.schemaTitle": "Search results for {searchTitle} | {marketplaceName}",
+  "SeatsFilter.labelSelected":"Seats - {seatCount}",
```
Gnito, line 758: This probably needs a bit of discussion with copy-text department. :)

We currently have 3 different ways to mark which filters are active. This would be 4th => I suggest that we'll stick to existing: `"Seats: {seatCount}"`.
Screenshot 2024-12-03 at 14 31 39

src/components/FieldNumber/FieldNumber.js
```diff
@@ -0,0 +1,95 @@
...
+  return (
+    <span className={css.numberInputWrapper}>
+      <button
+        className={css.numberButton}
+        name={decrement}
+        type="button"
+        onClick={handleValueChange}
+        disabled={value <= minValue}
+      >
+        -
```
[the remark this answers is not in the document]
Gnito: Yes. The plus and minus should be SVG icons.
You would also need to center them, which you usually can do with flexbox (and sometimes you might need to use a padding for fine-tuning the vertical position).

> _Inside a button it might be somewhat tricky to position svg & text correctly (browsers have their own layout rules for buttons, which makes life sometimes difficult). Anyway, it does not look like you need to do that here, but an old trick has been to add <span> inside the button which then wraps all the content there. That way, you ppl can control the display mode and centering a bit better._

src/translations/en.json
SariSaar: This translation is used on the map page and it's specified in the design:
Screenshot 2024-12-03 at 14 48 28

The grid page i.e. plain filter currently uses `FilterComponent.seatsLabel` like this:

```
    const labelForPlain = hasInitialValues ? (
      <span className={css.labelPlainSelected}>
        {label}: {initialSeats}
      </span>
    ) : (
      <span className={css.labelPlain}>{label}</span>
    );
```
but I can add a separate label translation for the plain version as well.

src/components/FieldNumber/FieldNumber.module.css
```diff
@@ -0,0 +1,46 @@
...
+
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  cursor: pointer;
+  min-width: 104px;
+}
+
+.numberButton {
+  border-width: thin;
```
Gnito, line 19: We have used pixels.

src/containers/SearchPage/SearchPage.shared.js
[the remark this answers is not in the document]
Gnito: Ah, right. Sorry, useless comment!

src/translations/en.json
Gnito: That probably makes sense. 

I think that our Integer range should actually behave differently than it currently does:
Screenshot 2024-12-03 at 15 24 20

I.e. it should look like this: `Gears: 18 - 24` since random range "18 - 24" does not specify which listing field filter it is + the range should fit on that button label concept.

I guess it could be achieved with just something like this: 

```jsx
  label={intl.formatMessage({ id: 'IntegerRangeFilter.labelSelectedPopup' }, {minValue, maxValue})}
  //   "IntegerRangeFilter.labelSelectedPopup": "{label}: {minValue} - {maxValue}",
```

src/components/FieldNumber/FieldNumber.js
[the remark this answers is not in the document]
SariSaar: Modified to SVGs

Screenshot 2024-12-03 at 17 46 54

src/components/FieldNumber/FieldNumber.module.css
SariSaar: When changing this to SVGs, border-width became stroke-width but I set that with pixels now 👍

src/containers/SearchPage/SearchPageWithGrid.js
SariSaar: I tested this and changed it to use mergedQueryParams for clarity 👍

src/containers/SearchPage/SeatsFilter/SeatsFilter.js
SariSaar: Default props removed and propTypes converted to JSDoc

src/translations/en.json
SariSaar: Added a new key `SeatsFilter.labelSelectedPlain` for the grid page version and renamed the popup key to `SeatsFilter.labelSelectedPopup` for clarity

src/components/FieldNumber/FieldNumber.js
```diff
@@ -0,0 +1,93 @@
...
+
+  return (
+    <span className={css.numberInputWrapper}>
+      <button
+        className={css.numberButton}
+        name={decrement}
+        type="button"
+        onClick={handleValueChange}
+        disabled={value <= minValue}
+      >
```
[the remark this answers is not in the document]
SariSaar: Added title tags to FieldNumber buttons

src/translations/en.json
SariSaar: Changed both versions so they use a single key, `SeatsFilter.labelSelected`, which has the colon option.

src/containers/SearchPage/SeatsFilter/SeatsFilter.js
```diff
@@ -0,0 +1,161 @@
...
+  id: string.isRequired,
+  name: string.isRequired,
+  queryParamNames: arrayOf(string).isRequired,
+  label: string.isRequired,
+  onSubmit: func.isRequired,
+  initialValues: shape({
+    seats: oneOfType([string, number]),
+  }),
+  contentPlacementOffset: number,
+ */
```
Gnito, line 34: I guess these should be removed.

PR #530, opened 2025-01-13
Update CustomExtendedDataField.module.css

AleksaButterfly:
Added the missing `customMediaQueries.css` import, which caused the viewport-based styles to be ineffective.


Gnito:
Thanks for the PR! I picked the commit to a separate branch and added a changelog entry.

PR #535, opened 2025-01-20
Fix: offer's schema should use dot as subunit separator

Gnito:
Let's not localize the price schema, but use Number with fixed number of decimal places.
https://schema.org/price


src/containers/ListingPage/ListingPage.shared.js
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { FormattedMessage } from '../../util/reactIntl';
+import { createIntl, FormattedMessage } from '../../util/reactIntl';
```
shareoc, line 2: Why is there a `createIntl` import?

Gnito: Good question! No need for that (I was testing another approach first)

PR #536, opened 2025-01-21
TransactionPage.duck.js: fix fetchMonthlyTimeSlots for inquiries

Gnito:
Inquiries didn't take unitType into account when fetching timeSlots for day & night bookings.
This caused current day to be available for the booking start time (which then jammed the end-time picker)

Note: current day is not selectable atm.


SariSaar:
LGTM!

PR #537, opened 2025-01-23
wip (private marketplace: sitemap crawling issue)

Gnito:
Suggestion to redirect /sitemap-... routes to /robots/sitemap-...

Before private marketplace feature, we had 2 resources: /s and /sitemap. 
Disallowing /s on robots.txt is the same as disallowing the path with a wildcard /s* (which basically disallows bots from following links to /sitemap-index.xml too)


Gnito:
Let’s use Allow: rule to get smaller footprint.

PR #539, opened 2025-01-23
add support for new google places api

shareoc:
This PR introduces support for the New Google Places API. While the Legacy Places API will continue to work, users will not be able to enable the Legacy Places API after **March 1st, 2025**. Read more in [Google's documentation](https://developers.google.com/maps/documentation/javascript/places-migration-overview).

We make a call to Google's API and based on the response we identify if the New API or Legacy API is in use. 

After March 1st, we'll remove support for the Legacy API in the codebase.

To use the New Places API, you'll need to enable it in Google's API settings.

### DIFF
--- a/src/components/LocationAutocompleteInput/GeocoderGoogleMapsNew.js
+++ b/src/components/LocationAutocompleteInput/GeocoderGoogleMapsNew.js
@@ -1,0 +1,116 @@
+import React from 'react';
+import classNames from 'classnames';
+import * as googleMapsUtil from '../../util/googleMaps';
+import { userLocation } from '../../util/maps';
+
+import css from './LocationAutocompleteInput.module.css';
+
+export const CURRENT_LOCATION_ID = 'current-location';
+
+// When displaying data from the Google Maps Places API, and
+// attribution is required next to the results.
+// See: https://developers.google.com/places/web-service/policies#powered
+export const GeocoderAttribution = props => {
+  const { rootClassName, className } = props;
+  const classes = classNames(rootClassName || css.poweredByGoogle, className);
+  return <div className={classes} />;
+};
+
+/**
+ * A forward geocoding (place name -> coordinates) implementation
+ * using the Google Maps Places API.
+ */
+class GeocoderGoogleMaps {
+  constructor() {
+    this.sessionToken = null;
+  }
+  getSessionToken() {
+    this.sessionToken =
+      this.sessionToken || new window.google.maps.places.AutocompleteSessionToken();
+    return this.sessionToken;
+  }
+
+  // Public API
+  //
+
+  /**
+   * Search places with the given name.
+   *
+   * @param {String} search query for place names
+   *
+   * @return {Promise<{ search: String, predictions: Array<Object>}>}
+   * results of the geocoding, should have the original search query
+   * and an array of predictions. The format of the predictions is
+   * only relevant for the `getPlaceDetails` function below.
+   */
+  getPlacePredictions(search, countryLimit) {
+    const limitCountriesMaybe = countryLimit
+      ? {
+          componentRestrictions: {
+            country: countryLimit,
+          },
+        }
+      : {};
+
+    return googleMapsUtil
+      .getPlacePredictionsNew(search, this.getSessionToken(), limitCountriesMaybe)
+      .then(results => {
+        return {
+          search,
+          predictions: results.predictions,
+        };
+      });
+  }
+
+  getPredictionId(prediction) {
+    if (prediction.predictionPlace) {
+      // default prediction defined above
+      return prediction.id;
+    }
+    return prediction.placePrediction.placeId;
+  }
+
+  /**
+   * Get the address text of the given prediction.
+   */
+  getPredictionAddress(prediction) {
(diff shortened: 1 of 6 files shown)


src/components/LocationAutocompleteInput/LocationAutocompleteInputImpl.js
```diff
@@ -161,6 +172,32 @@ class LocationAutocompleteInputImplementation extends Component {
 
   componentDidMount() {
     this._isMounted = true;
+
+    // Check whether the new version of the Google Places API is in use.
+    // The useNewGooglePlacesAPI property indicates whether the updated
+    // Google Places API is available and being utilized. If the useNewGooglePlacesAPI
+    // property is undefined, it means we haven't determined the API version yet.
+    if (typeof window !== 'undefined') {
```
Gnito, line 180: This is unnecessary!

The **componentDidMount** is executed after the component has mounted to DOM (aka it will have window since this function is run on browser side)

src/config/configMaps.js
```diff
@@ -45,9 +45,9 @@ export const search = {
   sortSearchByDistance: false,
 
   // Limit location autocomplete to a one or more countries
-  // using ISO 3166 alpha 2 country codes separated by commas.
+  // using CLDR two-character region codes, separated by commas.
```
Gnito, line 48: This change sounds wrong.
- **_ISO 3166 alpha 2_** country code is a standard. 
  (CLDR two-character region code is likely following that standard too)
- Mapbox supports the standard, so we should not change the config away from it:
  https://docs.mapbox.com/api/search/geocoding/#forward-geocoding-with-search-text-input

src/config/configMaps.js
```diff
@@ -45,9 +45,9 @@ export const search = {
...
   // If you want to limit the autocomplete, uncomment this value:
-  // countryLimit: ['AU'],
+  // countryLimit: ['us'],
```
Gnito, line 50: Don't change this.

Did you try capital letters? If they work, keep it as they are. If not, then those values, need to be converted to lower case before they are given to Google Maps.

src/util/googleMaps.js
```diff
@@ -55,6 +92,33 @@ export const getPlaceDetails = (placeId, sessionToken) =>
...
+
+    await place.fetchFields({ fields: fields });
+
+    return {
+      adress: place.formattedAddress,
+      origin: placeOriginNew(place),
+      bounds: placeBoundsNew(place),
+    };
+  } catch (error) {
+    console.error(`Could not get details for place id "${placeId}": `, error);
```
Gnito, line 118: I don't think we need to fill our logs with these error messages.

What you could do is to add this to the beginning of the file:
`const isDev = process.env.NODE_ENV === 'development';`

and conditionally print the error:

```diff
if (isDev) {
  console.error(`Could not get details for place id "${placeId}": `, error);
}
```

Gnito: Actually, since this is marked to be an async function. You should return a rejected promise:

```js
return Promise.reject(error);
```

src/components/LocationAutocompleteInput/GeocoderGoogleMapsNew.js
```diff
@@ -0,0 +1,116 @@
...
+          bounds: googleMapsUtil.locationBounds(latlng, currentLocationBoundsDistance),
+        };
+      });
+    }
+
+    if (prediction.predictionPlace) {
+      return Promise.resolve(prediction.predictionPlace);
+    }
+
+    return googleMapsUtil.getPlaceDetailsNew(this.getPredictionId(prediction)).then(place => {
```
Gnito, line 109: Since getPlaceDetailsNew is marked to be asynchronous, you need to change it so that it returns a Promise always - otherwise `.then` chaining will be called for undefined.

src/components/LocationAutocompleteInput/GeocoderGoogleMapsNew.js
```diff
@@ -0,0 +1,116 @@
...
+      return userLocation().then(latlng => {
+        return {
+          address: '',
+          origin: latlng,
+          bounds: googleMapsUtil.locationBounds(latlng, currentLocationBoundsDistance),
+        };
+      });
+    }
+
+    if (prediction.predictionPlace) {
```
Gnito, line 105: Did you test that everything works with the default location searches config?
https://github.com/sharetribe/web-template/blob/main/src/config/configDefaultLocationSearches.js#L15-L20

This predictionPlace is related to that setup.

src/components/LocationAutocompleteInput/GeocoderGoogleMapsNew.js
shareoc: Don't async functions return a Promise implicitly? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#description

shareoc: Or are you referring to the error handling

Gnito: Ah, right!

PR #542, opened 2025-01-27
Fix dark skin tab focus styles overriding hover font color

AleksaButterfly:
This commit resolves an issue where the tabContentDarkSkin class in TabNavHorizontal.module.css did not include :focus styles, causing it to inherit the :focus styles from the tabContent class. Since the tabContent class uses black font color and the dark skin uses white font color, this created a visibility issue when the component was in its dark skin mode and focused.


Gnito:
This is fixed in #543 (It's using the commit from this PR). Closing this one.

PR #561, opened 2025-03-13
Search as cta form

shareoc:
Introduces a search bar that can be toggled via assets. Uses dummy filter fields, with no actual functionality yet.


src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,120 @@
...
+          queryParams[key] = value;
+        }
+      }
+    });
+
+    // Create search string
+    const searchString = stringify(queryParams);
+
+    // Use history.push to navigate without page refresh
+    const to = createResourceLocatorString('SearchPage', routeConfiguration, {}, {});
```
Gnito, line 65: Check this function. It already has a handling for the search params ;)

src/translations/en.json
```diff
@@ -750,6 +750,7 @@
   "SavedCardDetails.replaceCardTitle": "Use another payment card",
   "SavedCardDetails.savedPaymentMethodPlaceholderDesktop": "•••• •••• •••• {last4Digits}",
   "SavedCardDetails.savedPaymentMethodPlaceholderMobile": "…{last4Digits}",
+  "SearchCTA.ButtonLabel": "Search",
```
Gnito, line 753:
```suggestion
  "SearchCTA.buttonLabel": "Search",
```

or   "SearchCTA.submitButtonLabel": "Search",

src/containers/PageBuilder/SectionBuilder/SectionContainer/SectionContainer.module.css
```diff
@@ -23,6 +23,8 @@
   padding: 32px 0;
   position: relative;
   word-break: break-word;
+  width: 100%;
+  /* also, TODO, see if this breaks anyting */
```
Gnito, line 27: This is a good question - since this affects every section type. Is this needed at all?

Another option would be to create some way for a sectionSectionBuilder to pass forward a prop or class that overrides or extends .sectionContent class.
- https://github.com/sharetribe/web-template/blob/main/src/containers/PageBuilder/SectionBuilder/SectionContainer/SectionContainer.js#L45
- https://github.com/sharetribe/web-template/blob/main/src/containers/PageBuilder/SectionBuilder/SectionBuilder.js#L129

That would allow passing a property that's only used if the _search CTA_ is in use.

However, I'm not sure if the extra complexity makes sense. The `width: 100%` is more straightforward solution.

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
...
+import { createResourceLocatorString } from '../../../../util/routes';
+import { useRouteConfiguration } from '../../../../context/routeConfigurationContext';
+import { useConfiguration } from '../../../../context/configurationContext';
+import { FormattedMessage } from '../../../../util/reactIntl';
+
+import FilterCategories from './FilterCategories';
+import FilterDateRange from './FilterDateRange';
+import FilterLocation from './FilterLocation';
+import FilterKeyword from './FilterKeyword';
+import { Button, Form } from '../../../../components';
```
Gnito, line 14: This should be before local files. I.e. import first shared components and only after that component from the same directory.

---

How is this affecting the created code chunks? How the size of the code chunks change between this branch and main branch? 
(I've tried to keep the PageBuilder stuff quite independent from the rest of the codebase (to avoid the need to download huge code chunks on the landing page. However, it might be that that battle is already lost because of topbar.)

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import classNames from 'classnames';
+import { Form as FinalForm } from 'react-final-form';
+import { useHistory } from 'react-router-dom';
+import { createResourceLocatorString } from '../../../../util/routes';
+import { useRouteConfiguration } from '../../../../context/routeConfigurationContext';
+import { useConfiguration } from '../../../../context/configurationContext';
+import { FormattedMessage } from '../../../../util/reactIntl';
```
Gnito, line 8:
```suggestion
// Contexts
import { useRouteConfiguration } from '../../../../context/routeConfigurationContext';
import { useConfiguration } from '../../../../context/configurationContext';
// Utils
import { FormattedMessage } from '../../../../util/reactIntl';
import { createResourceLocatorString } from '../../../../util/routes';
```

Perhaps it would be a bit cleaner and futureproof to keep the order of different imports constant. Like this:
https://github.com/sharetribe/web-template/blob/main/src/containers/ListingPage/ListingPageCarousel.js#L6-L12

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
...
+        queryParams[key] = value;
+      }
+    });
+
+    const to = createResourceLocatorString('SearchPage', routeConfiguration, {}, queryParams);
+    // Use history.push to navigate without page refresh
+    history.push(to);
+  };
+
+  const config = useConfiguration();
```
Gnito, line 64: Move this to the top (next to useHistory, useRouteConfiguration, etc.)

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
...
+    ) : null;
+  const locationMaybe = locationSearch ? (
+    <div className={css.filterField}>{/* <FilterLocation /> */}</div>
+  ) : null;
+  const keywordsMaybe = keywordSearch ? (
+    <div className={css.filterField}>{<FilterKeyword />}</div>
+  ) : null;
+  const dateRangeMaybe = dateRange ? (
+    <div className={css.filterField}>{/* <FilterDateRange /> */}</div>
+  ) : null;
```
Gnito, line 89: Another possibility to consider:

```suggestion
  const filterMap = {
    categories: { 
      shouldInclude: !!categories && categoryConfig.categories.length > 0, 
      component: <FilterCategories />,
    },
    locationSearch: { 
      shouldInclude: !!locationSearch, 
      component: <FilterLocation />,
    },
    keywordSearch: { 
      shouldInclude: !!keywordSearch, 
      component: <FilterLocation />,
    },
    dateRange: { 
      shouldInclude: !!dateRange, 
      component: <FilterDateRange />,
    },
  };
  const addFilterMaybe = key => {
    const { shouldInclude, component }  = filterMap[key] || {};
    return shouldInclude ? <div className={css.filterField}>{component}</div> : null;
  }
  
  // Usage (one by one):
  // {addFilterMaybe('categories')}
  // {addFilterMaybe('locationSearch')}
  // {addFilterMaybe('keywordSearch')}
  // {addFilterMaybe('dateRange')}
  
  // or oneliner:
  // {Object.keys(filterMap).map(k => addFilterMaybe(k))}
```

Just an idea to explore other ways to handle a this kind of structure. (Disclaimer: untested code)

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
...
+    categories && categoryConfig.categories.length > 0 ? (
+      <div className={css.filterField}>
+        {/* <FilterCategories categories={categoryConfig.categories}/> */}
+      </div>
+    ) : null;
+  const locationMaybe = locationSearch ? (
+    <div className={css.filterField}>{/* <FilterLocation /> */}</div>
+  ) : null;
+  const keywordsMaybe = keywordSearch ? (
+    <div className={css.filterField}>{<FilterKeyword />}</div>
```
Gnito, line 85:
```suggestion
    <div className={css.filterField}><FilterKeyword /></div>
```

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,117 @@
...
+
+  const config = useConfiguration();
+  const categoryConfig = config.categoryConfiguration;
+
+  // Count the number search fields that are enabled
+  const fieldCount = [categories, dateRange, keywordSearch, locationSearch].filter(field => !!field)
+    .length;
+
+  // conditional; even if categories are enabled, we won't render categories if none are present. So need to do a bit of calculations here
+  const fieldCountForGrid = categoryConfig.categories.length > 0 ? fieldCount : fieldCount - 1;
```
Gnito, line 72: How do you know if categories was one of the fields that contributed to fieldCount?

src/containers/PageBuilder/SectionBuilder/SectionBuilder.module.css
```diff
@@ -101,6 +101,9 @@
 
   @media (--viewportMedium) {
     justify-content: center;
+    grid-template-columns: 1fr;
+    place-items: center;
+    /* Let's see if this breaks anything TODO */
```
Gnito, line 106: Why these are needed?

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
Gnito: @shareoc you need to move this check inside the filter that is used to define the count of active filters.
`[categories, dateRange, keywordSearch, locationSearch].filter(field => !!field)`

```
  // Note: JS maintains the order of properties introduced to the object
  const filters = {
    categories: { enabled: categories , isValid: () => categoryConfig.categories.length > 0 },
    dateRange: { enabled: dateRange, isValid: () => true },
    keywordSearch: { enabled: keywordSearch, isValid: () => true },
    locationSearch: { enabled: locationSearch, isValid: () => true },
  };

  // Count the number search fields that are enabled
  const fieldCountForGrid = Object.values(filters).filter(field => field.enabled && field.isValid());
  }).length;
```
Disclaimer: untested code

src/translations/en.json
```diff
@@ -770,6 +772,11 @@
   "SavedCardDetails.replaceCardTitle": "Use another payment card",
   "SavedCardDetails.savedPaymentMethodPlaceholderDesktop": "•••• •••• •••• {last4Digits}",
   "SavedCardDetails.savedPaymentMethodPlaceholderMobile": "…{last4Digits}",
+  "SearchCTA.buttonLabel": "Search",
+  "SearchCTA.dateFilterPlaceholder": "Dates",
+  "SearchCTA.keywordFilterPlaceholder": "Search listings",
+  "SearchCTA.locationPlaceholder": "Location",
+  "SearchCTA.currentLocationPlaceholder": "Current location",
```
Gnito, line 779: Perhaps these should also be namespaced 
E.g. "Pages.SearchCTA.*" or "PageBuilder.SearchCTA.*"

src/components/IconDate/IconDate.js
```diff
@@ -0,0 +1,35 @@
...
+ * @param {Object} props
+ * @param {string?} props.className add more style rules in addition to components own css.root
+ * @param {string?} props.rootClassName overwrite components own css.root
+ * @returns {JSX.Element} SVG icon
+ */
+const IconDate = props => {
+  const { rootClassName, className } = props;
+  const classes = classNames(rootClassName || css.root, className);
+  return (
+    <svg
```
Gnito, line 19: Did you run this through SVGOMG?
https://jakearchibald.github.io/svgomg/

src/components/IconLocation/IconLocation.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {Object} props
+ * @param {string?} props.className add more style rules in addition to components own css.root
+ * @param {string?} props.rootClassName overwrite components own css.root
+ * @returns {JSX.Element} SVG icon
+ */
+const IconLocation = props => {
+  const { rootClassName, className } = props;
+  const classes = classNames(rootClassName || css.root, className);
+  return (
+    <svg
```
Gnito, line 19: Did you run this through SVGOMG
https://jakearchibald.github.io/svgomg/

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.js
```diff
@@ -0,0 +1,111 @@
...
+  return (
+    <OutsideClickHandler className={classes} onOutsideClick={() => setIsOpen(false)}>
+      <div className={css.dropdownContainer}>
+        <div
+          className={classNames(
+            css.toggleButton,
+            // If no selection has been made, no category is selected, and dropdown is closed, apply 'unselected' style
+            // else, if dropdown is open but no selection has been made, apply placeholderOpened style
+            !hasSelected && !selectedCategory && !isOpen && css.unselected,
+            isOpen && !hasSelected && css.placeholderOpened
```
Gnito, line 49:
```suggestion
            css.toggleButton,
            {
              // If no selection has been made, no category is selected, and dropdown is closed, apply 'unselected' style
              // else, if dropdown is open but no selection has been made, apply placeholderOpened style
              [css.unselected]: !hasSelected && !selectedCategory && !isOpen,
              [css.placeholderOpened]: isOpen && !hasSelected,
            }
```

Do I understand this right: 
When filter is open and there's no selected category nor "all values", the grey700 color is used => aka filter shows as active?

(I find this slightly counter-intuitive when compared to <input> behaviour: if there's no value, placeholder text is shown as grey => i.e. whenever placeholder text is visible, it's grey)

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.js
```diff
@@ -0,0 +1,111 @@
...
+          onClick={toggleDropdown}
+          type="button"
+        >
+          <span>{labelText}</span>
+          <span className={classNames(css.chevron, isOpen && css.isOpen)} />
+        </div>
+
+        {isOpen && (
+          <div className={css.dropdownContent}>
+            <div
```
Gnito, line 60: <button>

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.js
```diff
@@ -0,0 +1,111 @@
...
+                  hasSelected && input.value === ''
+                    ? css.dropdownItemBorderSelected
+                    : css.dropdownItemBorder
+                }
+              />
+              <FormattedMessage id="CategoryFilter.selectAll" />
+            </div>
+
+            {categories.map(({ id, name }) => (
+              <div
```
Gnito, line 76: <button>

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.module.css
```diff
@@ -0,0 +1,146 @@
...
+  background-color: var(--colorGrey900);
+  width: 6px;
+}
+
+.chevron {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  transition: transform 0.4s;
+  -webkit-transition: -webkit-transform 0.4s;
```
Gnito, line 113: Our build setup should add these browser engine prefixes if needed. `-webkit-*`
Remove this and check that transition still happens on Safari/Chrome.

src/containers/PageBuilder/Primitives/SearchCTA/FilterDateRange/FilterDateRange.module.css
```diff
@@ -0,0 +1,69 @@
...
+    margin-top: 0;
+  }
+}
+
+.toggleButton {
+  display: flex;
+  align-items: center;
+  height: 40px;
+  padding: 0 12px;
+  border-radius: 4px;
```
Gnito, line 39: --borderRadiusMedium

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.module.css
```diff
@@ -0,0 +1,146 @@
...
+    padding: 0;
+  }
+}
+
+.toggleButton {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  height: 40px;
+  border-radius: 4px;
```
Gnito, line 30: --borderRadiusMedium

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.module.css
```diff
@@ -0,0 +1,123 @@
+@import '../../../../styles/customMediaQueries.css';
+
+.filterField {
+  height: 100%;
+  align-items: center;
+  display: grid;
+  min-height: 40px;
```
Gnito, line 7: Try to keep related CSS values next to each other.

```css
  display: grid; /* display model is grid */
  align-items: center; /* grid should align its items to the center*/
  height: 100%; /* height in general*/
  min-height: 40px; /* an edge case for height below it */
```

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.module.css
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.align {
+  justify-self: start;
+  @media (--viewportMedium) {
+    justify-self: center;
+  }
+}
+
+.searchBarContainer {
```
Gnito, line 23: You have 3 places where you touch .searchBarContainer (1 for each breakpoint). However, you also have a nested breakpoint inside the scope of this first ruleset.

Even though it's perfectly valid way to add breakpoints and add multiple rulesets for a class (nested inside media queries), I'd suggest that you do the opposite: nest all the breakpoints under one class.

The reason is that over the years it's easier to manage the ruleset of a single class, when you see all the CSS declarations of different breakpoint versions with a single glance. I.e. you make change to a desktop layout and immediately see related declarations for mobile / table scopes. It's less likely that future me/you forgets to adjust all the different values accordingly. (In addition, media query nesting has already landed to most browsers - so, it's standard CSS nowadays.)

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.module.css
```diff
@@ -0,0 +1,123 @@
...
+  composes: align;
+  background-color: var(--colorWhite);
+  border-color: var(--colorGrey800);
+  border-width: 1px;
+  color: var(--colorGrey500);
+  border-radius: 8px;
+  box-shadow: var(--boxShadowPopup);
+  width: 100%;
```
Gnito, line 31: Something like this perhaps:
```suggestion
  width: calc(100vw - 72px);
```
or `width: 80vw,` but this is not as accurate with the margins.

If the value is "100%", the item inside a grid is taking the content width instead of grid's width.

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.module.css
```diff
@@ -0,0 +1,123 @@
...
+    padding: 8px;
+  }
+}
+
+.gridContainer {
+  height: 100%;
+  display: grid;
+  align-items: center;
+  justify-content: center;
+  align-items: center;
```
Gnito, line 43: Don't add this twice

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.module.css
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.gridContainer > *:not(:last-child):not(:nth-last-child(2)) {
+  border-bottom: 1px solid var(--colorGrey200);
+}
```
Gnito, line 48:
```suggestion
.gridContainer {
  height: 100%;
  display: grid;
  align-items: center;
  justify-content: center;

  & > *:not(:last-child):not(:nth-last-child(2)) {
    border-bottom: 1px solid var(--colorGrey200);
  }

  @media (--viewportLarge) {
    padding-top: 0;

    & > *:not(:last-child):not(:nth-last-child(2)) {
      border-right: 1px solid var(--colorGrey200);
      border-bottom: none;
      padding-bottom: 0;
    }

    & > *:last-child {
      justify-self: end;
    }
  }
}
```

src/containers/PageBuilder/Primitives/SearchCTA/SearchCTA.js
```diff
@@ -0,0 +1,176 @@
...
+    .length;
+
+  //  If no search fields are enabled, we return null (Console won't allow you to enable 0 search fields)
+  if (!fieldCountForGrid) {
+    return null;
+  }
+
+  const onSubmit = values => {
+    // Convert form values to query parameters
+    const queryParams = {};
```
Gnito, line 118:
```suggestion
    let queryParams = {};
```
This is not a constant, so [let](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) is more appropriate.

JavaScript's **_const_** does not block a re-assignment to nested properties of an object, but it's still good to be transparent what's the use case.

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.module.css
```diff
@@ -0,0 +1,146 @@
...
+    left: -8px;
+    transform: none;
+    width: 100%;
+  }
+}
+
+.option {
+  position: relative;
+  width: 100%;
+  padding: 7px 0 7px 22px;
```
Gnito, line 69: There needs to be padding also on the right-hand side.
Screenshot 2025-04-23 at 21 58 07

src/containers/PageBuilder/Primitives/SearchCTA/FilterDateRange/FilterDateRange.js
```diff
@@ -0,0 +1,80 @@
...
+  const [selectedDates, setSelectedDates] = useState(null);
+  const { className, rootClassName, config } = props;
+  const intl = useIntl();
+
+  const classes = classNames(rootClassName || css.root, className);
+
+  const formatDateRange = (start, end) => {
+    const formattedStart = intl.formatDate(start, { day: 'numeric', month: 'short' });
+    const formattedEnd = intl.formatDate(end, { day: 'numeric', month: 'short' });
+    return `${formattedStart} - ${formattedEnd}`;
```
Gnito, line 31: I don't think we can use this format
Screenshot 2025-04-23 at 22 27 45

I guess the options are 
- set both day & month to numeric, but format them separately
  - fi: **10.1. - 12.1.**, en-US: **1/10 - 1/12**, en-GB: **10/01 - 12/01**
- Use [formatDateTimeRange](https://formatjs.github.io/docs/react-intl/api#formatdatetimerange)
  - This should look like this with month set to 'short': 
    fi: **10.–12. tammikuuta**, en-US: **Jan 10 – 12**, en-GB: **10–12 Jan**
  - or as both numeric: 
    fi: **10.–12.1.**, en-US: **1/10 – 1/12**, en-GB: **10/01 – 12/01**

I think all of these should fit to the space we have available.

Gnito: @shareoc perhaps you could escalate this to product channel and ping e.g. juho there?
I'm leaning towards second option (fi: 10.–12. tammikuuta, en-US: Jan 10 – 12, en-GB: 10–12 Jan)

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.js
shareoc: @Gnito placeholder should be shown grey700 when the dropdown is active:

image

src/containers/PageBuilder/Primitives/SearchCTA/FilterDateRange/FilterDateRange.js
Gnito: @Kusti what do you think?

Kusti: @shareoc @Gnito the second option looks like an obvious choice to me too, let's go with that!

src/containers/PageBuilder/Primitives/SearchCTA/FilterCategories/FilterCategories.js
```diff
@@ -7,86 +7,135 @@ import { OutsideClickHandler } from '../../../../../components';
...
+            })}
+            role="listbox"
+            id="category-listbox"
+          >
+            {allOptions.map(({ id, name }, index) => {
+              const isSelected =
+                id === input.value || (id === 'all-categories' && input.value === '');
+              const isActive = index === activeIndex;
+              return (
+                <div
```
Gnito, line 122: Ideally, these should be <li> elements and the parent should be <ul>

PR #565, opened 2025-03-24
replace zoomControlOptions with cameraControlOptions

shareoc:
By default, zoom is hidden, and instead we overlay Map camera controls, i.e. `cameraControlOptions`:

image

Therefore, `zoomControlOptions` has no effect on positioning, as zoom controls are hidden by default. 

`cameraControlOptions` is the correct attribute to use to position the cameraControlOptions.

An alternative here would be to continue using 

```
 zoomControlOptions: {
      position: controlPosition,
    },
```

and add

`zoom: true`

to the mapConfig object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@ way to update this template, but currently, we follow a pattern:
 ---
 
 ## Upcoming version 2025-XX-XX
+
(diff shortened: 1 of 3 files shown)


Gnito (approved):
LGTM. Add a changelog entry

PR #583, opened 2025-04-16
Listing type filtering

SariSaar:


src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -495,3 +533,24 @@ export const getDatesAndSeatsMaybe = (currentParams, newParams) => {
...
+ *    pathParams – either { listingType } or {},
+ *    search,
+ *
+ * ]
+ */
+export const getResourceLocatorStringParams = (routeConfiguration, search, listingType) => {
+  const routeName = listingType ? 'SearchPageWithListingType' : 'SearchPage';
+  const pathParams = listingType ? { listingType } : {};
+  return [routeName, routeConfiguration, pathParams, search];
+};
```
Gnito, line 556: Perhaps you could go to this kind of direction
```suggestion
// Note: pathName could be picked from React Router's location object
// const location = useLocation();
// location?.pathname
export const getResourceLocatorStringParams = (routeConfiguration, pathParams, search, pathName) => {
      const testURL = new URL(`http://my.marketplace.com${pathname}`);
      const matchedRoutes = matchPathname(testURL.pathname, routeConfiguration);
      const routeName = matchedRoutes?.[0]?.found?.route?.name || 'SearchPage';

      return [routeName, routeConfiguration, pathParams, search];
};
```

src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -136,6 +150,12 @@ const validURLParamForCategoryData = (prefix, categories, level, params) => {
     : {};
 };
 
+const validURLParamForListingTypeData = (listingTypes, param) => {
```
Gnito, line 153: Side note: in JS it's usually good to introduce and define variables before they are used. In this scenario (where this is called omitLimitedListingFieldParams function), it does not matter - but it's good rule of thumb to avoid hoisting problems. 
- https://developer.mozilla.org/en-US/docs/Glossary/Hoisting
- https://www.freecodecamp.org/news/what-is-hoisting-in-javascript-3/

(Just mentioning this, because I'm so used to define helpers (constants, vars, and functions) first and only below them have the function or component that uses them.)

SariSaar: Good catch! I added the listing type version of this function right under the existing category version – would it make sense to move both of them to the top of the file, since they're both used in the omitLimitedListingFieldParams function, or should we leave it as it is? 

I think it makes sense to have both of those functions in one place, and I'm wondering if moving the category version would mean that customizers would get into trouble if they've modified the existing category function.

src/containers/SearchPage/SearchPage.shared.js
SariSaar: Updated the function to take routeConfiguration and location as params and to return routeName and pathParams fetched with matchedRoutes!

jayenashar:
any instructions on enabling this before it's in the console?  this is what i did:

```diff
diff --git a/src/config/configSearch.js b/src/config/configSearch.js
index 0d835745e..5c0e2d11d 100644
--- a/src/config/configSearch.js
+++ b/src/config/configSearch.js
@@ -89,3 +89,9 @@ export const sortConfig = {
     },
   ],
 };
+
+export const listingTypeFilter = {
+  key: 'listingType',
+  enabled: true,
+  schemaType: 'listingType',
+};
diff --git a/src/util/configHelpers.js b/src/util/configHelpers.js
index 08ad85122..c3803a37e 100644
--- a/src/util/configHelpers.js
+++ b/src/util/configHelpers.js
@@ -1302,6 +1302,7 @@ const mergeSearchConfig = (
   categoryConfiguration,
   listingTypeConfig
 ) => {
+  hostedSearchConfig.listingTypeFilter ??= defaultSearchConfig.listingTypeFilter;
   // The sortConfig is not yet configurable through Console / hosted assets,
   // but other default search configs come from hosted assets
   const searchConfig = hostedSearchConfig?.mainSearch
```

src/containers/SearchPage/SearchPage.shared.js
Gnito: @SariSaar 
Ah right! This is not causing bugs, but I'd still suggest moving both above omitLimitedListingFieldParams. If customizers have extended validURLParamForCategoryData (which is somewhat unlikely), they probably want to check this sibling function too to ensure that both functions work as they intended.

SariSaar:
> any instructions on enabling this before it's in the console? this is what i did:
> 
> ```diff
> diff --git a/src/config/configSearch.js b/src/config/configSearch.js
> index 0d835745e..5c0e2d11d 100644
> --- a/src/config/configSearch.js
> +++ b/src/config/configSearch.js
> @@ -89,3 +89,9 @@ export const sortConfig = {
>      },
>    ],
>  };
> +
> +export const listingTypeFilter = {
> +  key: 'listingType',
> +  enabled: true,
> +  schemaType: 'listingType',
> +};
> diff --git a/src/util/configHelpers.js b/src/util/configHelpers.js
> index 08ad85122..c3803a37e 100644
> --- a/src/util/configHelpers.js
> +++ b/src/util/configHelpers.js
> @@ -1302,6 +1302,7 @@ const mergeSearchConfig = (
>    categoryConfiguration,
>    listingTypeConfig
>  ) => {
> +  hostedSearchConfig.listingTypeFilter ??= defaultSearchConfig.listingTypeFilter;
>    // The sortConfig is not yet configurable through Console / hosted assets,
>    // but other default search configs come from hosted assets
>    const searchConfig = hostedSearchConfig?.mainSearch
> ```

@jayenashar that's one approach, sure! I'll add the local config option to this pull request, too.

It's good to note that before a feature branch has been merged to main, it is very much subject to change. We heavily use interactive rebases during feature development so that the main branch commit history consists of clear, well-boundaried commits – and this means that both code and especially hashes in feature branch commits will keep changing up until the feature is merged. So you can already start using a feature from a feature branch, as long as you are prepared to do the necessary git merge work when the feature is actually released to the codebase.

jayenashar:
yeah i already had a discussion with vesa some time about about rebasing of feature branches.  i don't do it for my projects but yeah, i'll worry about it when i need to upgrade this project to a newer template.

src/containers/SearchPage/SearchPage.shared.js
SariSaar: Moved! ✅

src/containers/SearchPage/SearchPage.duck.js
```diff
@@ -114,13 +114,19 @@ export const searchListings = (searchParams, config) => (dispatch, getState, sdk
...
     return config.listing.enforceValidListingType
       ? {
           pub_listingType: listingTypes.map(l => l.listingType),
           // pub_transactionProcessAlias: listingTypes.map(l => l.transactionType.alias),
           // pub_unitType: listingTypes.map(l => l.transactionType.unitType),
         }
+      : isListingTypeVariant && isValidListingTypeParam
+      ? {
+          pub_listingType: listingTypePathParam,
+        }
```
Gnito, line 129: I think this needs to be swapped.
1. "isListingTypeVariant && isValidListingTypeParam" => a single valid listing type in use
2. config.listing.enforceValidListingType => the previous behaviour

```suggestion
    return isListingTypeVariant && isValidListingTypeParam
      ? {
          pub_listingType: listingTypePathParam,
        }
      : config.listing.enforceValidListingType
      ? {
          pub_listingType: listingTypes.map(l => l.listingType),
          // pub_transactionProcessAlias: listingTypes.map(l => l.transactionType.alias),
          // pub_unitType: listingTypes.map(l => l.transactionType.unitType),
        }
```

jayenashar:
i think the roadmap needs an update 😛 
image

src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -496,3 +538,35 @@ export const getDatesAndSeatsMaybe = (currentParams, newParams) => {
...
+ * @param {*} routes current route configuration
+ * @param {*} location current ReactRouter location
+ * @returns an object with the attributes routeName and pathParams, which can then be passed
+ * as the corresponding parameters to createResourceLocatorString
+ */
+export const getResourceLocatorStringParams = (routes, location) => {
+  const testURL = new URL(`http://my.marketplace.com${location.pathname}`);
+  const matchedRoutes = matchPathname(testURL.pathname, routes);
+
+  if (matchedRoutes.length > 0) {
```
Gnito, line 554: What should happen in unhappy path? 
Return nulls, throw error, default to landing page ?

Gnito (approved):

PR #590, opened 2025-04-29
Convert non-string category URL param to string

SariSaar:
This change fixes an issue where a category with a numeric key doesn't get highlighted in the grid search page filter. The URL param is interpreted as a numeric value, so it doesn't match the category key string. Enforcing the non-string value to a string ensures that numeric (and boolean) category keys are highlighted correctly.


Gnito (approved):
LGTM!
(Add changelog entry)

PR #598, opened 2025-05-02
Pass email from login to forgot password

Sandeep-Petwal:
When a user clicks "**Reset password**" on the login page, the email address they have already typed (if any) is now correctly passed to the forgot password page.


Sandeep-Petwal:
thank you

PR #605, opened 2025-05-06
Fix multiple search page routes

SariSaar:
This update addresses a few bugs related to the listing page filtering feature:
- In mobile search filter view, clicking "cancel" now maintains listing type path parameter
- When showing LimitedAccessBanner, exclude both search page versions
- With the configuration "only show top bar search input on search page", include both search page versions
- If a listing field is restricted to multiple valid listing types, show that field on listing type specific search pages corresponding to any of those listing types

In addition, this update changes the way the listing type path parameter is passed to helper functions, making it possible to pass other path parameters more easily in the future.

### DIFF
--- a/src/containers/SearchPage/SearchPage.shared.js
+++ b/src/containers/SearchPage/SearchPage.shared.js
@@ -51,17 +51,20 @@ export const omitLimitedListingFieldParams = (searchParams, filterConfigs) => {
     defaultFiltersConfig,
     listingCategories,
     activeListingTypes,
-    listingTypePathParam,
+    currentPathParams = {},
   } = filterConfigs;
+
+  const { listingType: listingTypePathParam } = currentPathParams;
   const categorySearchConfig = defaultFiltersConfig.find(f => f.schemaType === 'category');
   const listingTypeSearchConfig = defaultFiltersConfig.find(f => f.schemaType === 'listingType');
   const validNestedCategoryParamNames = categorySearchConfig
     ? validURLParamForCategoryData(categorySearchConfig.key, listingCategories, 1, searchParams)
     : {};
 
-  const validListingTypeParamNames = listingTypeSearchConfig
-    ? validURLParamForListingTypeData(activeListingTypes, searchParams)
-    : {};
+  const validListingTypeParamNames =
+    activeListingTypes && listingTypeSearchConfig
+      ? validURLParamForListingTypeData(activeListingTypes, searchParams)
+      : {};
 
   return Object.entries(searchParams).reduce((picked, searchParam) => {
     const [searchParamKey, searchParamValue] = searchParam;
@@ -127,6 +130,8 @@ export const validURLParamForExtendedData = (
     return hasValidDates ? { [queryParamName]: paramValue } : {};
   } else if (queryParamName === 'seats') {
     return paramValue ? { [queryParamName]: paramValue } : {};
+  } else if (queryParamName === 'pub_listingType') {
+    return paramValue.length > 0 ? { [queryParamName]: paramValue } : {};
   }
 
   // Resolve configurations for extended data filters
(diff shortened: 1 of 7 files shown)


Gnito (approved):
LGTM

src/containers/SearchPage/SearchPage.shared.js
```diff
@@ -45,23 +45,30 @@ const validURLParamForListingTypeData = (listingTypes, param) => {
...
+  const listingTypeParamMaybe = listingTypePathParam
+    ? { pub_listingType: listingTypePathParam }
     : {};
+  const validListingTypeParamNames =
+    activeListingTypes && listingTypeSearchConfig
+      ? validURLParamForListingTypeData(activeListingTypes, {
+          ...searchParams,
+          ...listingTypeParamMaybe,
+        })
+      : {};
```
Gnito, line 71: I think you should do this earlier!

E.g. in SearchPageWithMap:
```js
        return {
          currentQueryParams: omitLimitedListingFieldParams(
            {
              ...mergedQueryParams,
              ...updatedURLParams,
              ...(listingTypePathParam ? { pub_listingType: listingTypePathParam } : {}),
              ...keywordsMaybe,
              ...datesAndSeatsMaybe,
              address,
              bounds,
            },
            filterConfigs,
            currentPathParams
          ),
```

src/containers/SearchPage/SearchPageWithMap.js
```diff
@@ -411,11 +429,16 @@ export class SearchPageComponent extends Component {
...
+      ? validFilterParams(
+          validQueryParams,
+          {
+            listingFieldsConfig: customSecondaryFilters,
+            defaultFiltersConfig: [],
+            listingCategories,
+            activeListingTypes,
+          },
+          currentPathParams
+        )
```
Gnito, line 441: This feels weird to me. There are 2 different parameters, but they are not next to each other. 
`validFilterParams(validQueryParams, currentPathParams, filterConfigs)`
`validFilterParams(validQueryParams, currentPathParams, filterConfigs, options)`


This could perhaps even keep the original signature:
```suggestion
      ? validFilterParams(
          {
            ...validQueryParams,
            ...(listingTypePathParam ? { pub_listingType: listingTypePathParam } : {}),
          },
          {
            listingFieldsConfig: customSecondaryFilters,
            defaultFiltersConfig: [],
            listingCategories,
            activeListingTypes,
          })
```

Note: you have used the same concept inside **_omitLimitedListingFieldParams_** (SearchPage.shared.js):
Screenshot 2025-05-08 at 12 26 17

src/containers/SearchPage/SearchPage.shared.js
Gnito: I.e. omitLimitedListingFieldParams would keep the original parameter signature - and pub_listingType is enforced to be the one that's found from the path parameters.

SariSaar: I tested this now, and with this change the pub_listingType param gets passed to the URL even when the listingTypePathParam is defined i.e. when it shouldn't be included. 

So it turns out the omitLimitedListingFieldParams actually does need the information on whether this is a search param or a path param – it returns the search params that get set to the URL, and we don't want to set pub_listingType to URL in all cases.

src/containers/SearchPage/SearchPageWithMap.js
SariSaar: Had an IRL conversation about this, and we decided to pass currentPathParams inside filterConfig

src/containers/SearchPage/SearchPage.shared.js
SariSaar: Had a conversation on this IRL, and decided to pass currentPathParams in filterConfig to these functions

Gnito (approved):
LGTM

PR #608, opened 2025-05-15
Fix issue in emptying image array

SariSaar:
EditListingPhotosForm: removing all images from a listing caused the screen to go blank


Gnito (approved):
LGTM

PR #612, opened 2025-05-20
fix an issue with anchor links on modals on the signup/login page

shareoc:
This PR fixes an issue observed at least in Safari v18.4 where anchor links on the Privacy Policy and Terms of Service pages did not scroll to the correct section when those pages were opened in modals (e.g., via links in the sign-up or log-in forms).  The update ensures that anchor links now reliably scroll to the intended section.

### DIFF
--- a/src/containers/PrivacyPolicyPage/PrivacyPolicyPage.js
+++ b/src/containers/PrivacyPolicyPage/PrivacyPolicyPage.js
@@ -26,16 +26,17 @@ const SectionBuilder = loadable(
 const PrivacyPolicyContent = props => {
   const { inProgress, error, data } = props;
 
-  if (inProgress) {
-    return null;
-  }
-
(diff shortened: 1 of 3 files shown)


src/containers/PrivacyPolicyPage/PrivacyPolicyPage.js
```diff
@@ -26,16 +26,16 @@ const SectionBuilder = loadable(
...
-    return null;
-  }
-
   // We don't want to add h1 heading twice to the HTML (SEO issue).
   // Modal's header is mapped as h2
   const hasContent = data => typeof data?.content === 'string';
   const exposeContentAsChildren = data => {
     return hasContent(data) ? { children: data.content } : {};
   };
+
```
Gnito, line 35: Make the same change also to the TermsOfServicePage
(and add changelog entry)

Gnito (approved):
LGTM

PR #614, opened 2025-05-20
Add conditional visibility features based on user type definitions

SariSaar:
This change introduces conditional visibility on key elements depending on upcoming Console configurations for user types and top bar. This configuration is not yet released in Console!

Authenticated users with no user type, or with a user type that does not have any restrictions specified, will continue to see all of the following elements without changes.

## Visibility for "Create listings" and "Manage listings" links

Operators can hide the link for creating new listings
- from unauthenticated users
- and from specific user types

When the link configured as hidden, the following places will not show a link to create a listing
- the desktop top bar
- the mobile menu top bar
- any empty search results page

topbar_no_listinglink

In addition, an authenticated user will not see the Manage listings links in the profile menu or the navigation bar for profile and account settings.

profile_menu

usernav

## Payment details links in Account settings

Operators can select whether users with a specific user type can see tabs for adding payment methods, payout details, or both. The account settings side navigation shows the links configured for the user. Both payment method and payout detail pages will still be available for all authenticated users with a direct URL (`/payment-methods` or `/payments`).

account_sidenav

## Inbox links

Operators can configure user types to customer, provider, or both.

If a user's user type is defined as 'customer' only, the links across the template will direct that user to the `/orders` tab when visiting the inbox, and the side navigation will only show the customer tab option. The user can navigate to the `/sales` tab with a direct URL. 

Conversely, if a user's user type is defined as 'provider' only, the user will be linked to the `/sales` tab when visiting the inbox. They can navigate to the `/orders` tab with a direct URL. A user with both user type roles will continue to see both tabs.

provider_inbox

## Profile page review tabs

If a user's user type is defined as 'customer' only, their profile page will show a tab with their reviews as a customer in transactions. For a user type defined as 'provider' only, their profile page will show a tab with their reviews from customers. A user with both definitions, or with no definition, will have both kinds of reviews visible in their profile.

reviews_as_customer

### DIFF
--- a/src/components/LayoutComposer/LayoutSideNavigation/LayoutSideNavigation.js
+++ b/src/components/LayoutComposer/LayoutSideNavigation/LayoutSideNavigation.js
@@ -36,7 +36,7 @@ const LayoutSideNavigation = props => {
     footer: footerContent,
     sideNav: sideNavContent,
     useAccountSettingsNav,
-    currentPage,
+    accountSettingsNavProps,
     ...rest
   } = props;
 
@@ -63,7 +63,9 @@ const LayoutSideNavigation = props => {
             <Main as="div" className={containerClasses}>
               <aside className={classNames(css.sideNav, sideNavClassName)}>
                 {useAccountSettingsNav ? (
-                  <LayoutWrapperAccountSettingsSideNav currentPage={currentPage} />
+                  <LayoutWrapperAccountSettingsSideNav
+                    accountSettingsNavProps={accountSettingsNavProps}
+                  />
                 ) : null}
                 {sideNavContent}
               </aside>
--- a/src/components/MenuContent/MenuContent.js
+++ b/src/components/MenuContent/MenuContent.js
@@ -50,10 +50,10 @@ const MenuContent = props => {
   ) : null;
 
   React.Children.forEach(children, child => {
-    if (child.type !== MenuItem) {
+    if (child && child.type !== MenuItem) {
       throw new Error('All children of MenuContent must be MenuItems.');
     }
-    if (child.key == null) {
+    if (child && child.key == null) {
       throw new Error('All children of MenuContent must have a "key" prop.');
     }
   });
--- a/src/containers/TopbarContainer/Topbar/TopbarDesktop/CustomLinksMenu/CustomLinksMenu.js
+++ b/src/containers/TopbarContainer/Topbar/TopbarDesktop/CustomLinksMenu/CustomLinksMenu.js
@@ -6,16 +6,21 @@ import LinksMenu from './LinksMenu';
 import css from './CustomLinksMenu.module.css';
 
 const draftId = '00000000-0000-0000-0000-000000000000';
-const createListingLinkConfig = intl => ({
-  group: 'primary',
-  text: intl.formatMessage({ id: 'TopbarDesktop.createListing' }),
-  type: 'internal',
-  route: {
-    name: 'EditListingPage',
-    params: { slug: 'draft', id: draftId, type: 'new', tab: 'details' },
-  },
-  highlight: true,
-});
+const createListingLinkConfigMaybe = (intl, showLink) =>
+  showLink
+    ? [
+        {
+          group: 'primary',
+          text: intl.formatMessage({ id: 'TopbarDesktop.createListing' }),
+          type: 'internal',
+          route: {
+            name: 'EditListingPage',
+            params: { slug: 'draft', id: draftId, type: 'new', tab: 'details' },
+          },
+          highlight: true,
+        },
+      ]
+    : [];
 
 /**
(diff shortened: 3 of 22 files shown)


src/components/LayoutComposer/LayoutSideNavigation/LayoutSideNavigation.js
```diff
@@ -63,7 +65,11 @@ const LayoutSideNavigation = props => {
             <Main as="div" className={containerClasses}>
               <aside className={classNames(css.sideNav, sideNavClassName)}>
                 {useAccountSettingsNav ? (
-                  <LayoutWrapperAccountSettingsSideNav currentPage={currentPage} />
+                  <LayoutWrapperAccountSettingsSideNav
+                    currentPage={currentPage}
+                    showPaymentMethods={showPaymentMethods}
+                    showPayoutDetails={showPayoutDetails}
+                  />
                 ) : null}
```
Gnito, line 73: After live discussion: Let's create **_accountSettingsNavProps_** wrapper that passes all these props through the in-between components.

(LayoutWrapperAccountSettingsSideNav has translation keys, which makes it quite significant change to move this to come entirely through **_sideNavContent_** prop)

SariSaar: Updated to use **accountSettingsNavProps**!

src/containers/TopbarContainer/Topbar/TopbarDesktop/CustomLinksMenu/CustomLinksMenu.js
```diff
@@ -100,12 +103,23 @@ const calculateContainerWidth = (containerRefTarget, parentWidth) => {
...
+  intl,
+  showCreateListingsLink,
+}) => {
   const containerRef = useRef(null);
   const observer = useRef(null);
   const [mounted, setMounted] = useState(false);
   const [moreLabelWidth, setMoreLabelWidth] = useState(0);
-  const [links, setLinks] = useState([createListingLinkConfig(intl), ...customLinks]);
+  // TODO think about whether there's a better way to handle showCreateListingsLink
+  // TODO test SSR edge cases
```
Gnito, line 118: Does these TODOs still need attention?

src/containers/TopbarContainer/Topbar/TopbarDesktop/TopbarDesktop.js
```diff
@@ -55,28 +51,30 @@ const InboxLink = ({ notificationCount, currentUserHasListings }) => {
...
-        <MenuItem key="ManageListingsPage">
-          <NamedLink
-            className={classNames(css.menuLink, currentPageClass('ManageListingsPage'))}
-            name="ManageListingsPage"
-          >
-            <span className={css.menuItemBorder} />
-            <FormattedMessage id="TopbarDesktop.yourListingsLink" />
-          </NamedLink>
-        </MenuItem>
+        <MenuItem key="ManageListingsPage">{manageListingsLinkMaybe}</MenuItem>
```
Gnito, line 77: It sounds like this creates empty MenuItesm (`<li>` elements) to the menu:
`<MenuItem key="ManageListingsPage">{null}</MenuItem>`

Does this cause errors if you create the complete MenuItem conditionally?
```jsx
{showManageListingsLink ? (
  <MenuItem key="ManageListingsPage">
    <NamedLink
        className={classNames(css.menuLink, currentPageClass('ManageListingsPage'))}
        name="ManageListingsPage"
      >
        <span className={css.menuItemBorder} />
        <FormattedMessage id="TopbarDesktop.yourListingsLink" />
      </NamedLink>
  </MenuItem>
) : null}
```

I.e. I wonder if the null ends up here: 
https://github.com/sharetribe/web-template-private/blob/main/src/components/MenuContent/MenuContent.js#L53

SariSaar: Yes, creating the complete MenuItem conditionally breaks things because of that `child.type` check.

SariSaar: Updated the child checks in MenuContent to allow null and undefined children

shareoc (approved):

src/components/MenuContent/MenuContent.js
```diff
@@ -50,10 +50,10 @@ const MenuContent = props => {
   ) : null;
 
   React.Children.forEach(children, child => {
-    if (child.type !== MenuItem) {
+    if (child != null && child.type !== MenuItem) {
       throw new Error('All children of MenuContent must be MenuItems.');
     }
-    if (child.key == null) {
+    if (child != null && child.key == null) {
```
shareoc, line 56:
```suggestion
    if (child && child.key == null) {
```

PR #615, opened 2025-05-22
Add missing location prop to grid search page

SariSaar:
SearchPageWithGrid component wasn't passing location as a prop to SearchFiltersMobile, so the "Cancel" button was not working.


Gnito (approved):
LGTM

PR #616, opened 2025-05-22
Update a copy text for the Search as a CTA

thomasmalbaux:


thomasmalbaux:
@shareoc @Gnito The `PageBuilder.CategoryFilter.selectAll` key (and `PageBuilder.CategoryFilter.placeholder`) does not mention `SearchCTA` in its name, as other keys for the feature do.

But it is only used for the Search CTA, right? Or it is used elsewhere more generically?
Out of curiosity, what would be the reason why it would not be named `PageBuilder.SearchCTA.CategoryFilter.selectAll`?

PR #671, opened 2025-09-24
CheckoutPage: refactor how location is showed

Gnito:
Let's rely on location info saved to extended data. If it's not there, listing was created without that info - no need to show location section.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2025-XX-XX
 
+- [fix] CheckoutPageWithPayment: don't show location if it's unknown for bookings.
+  [#671](https://github.com/sharetribe/web-template/pull/671)
 - [add] Update README about translation files.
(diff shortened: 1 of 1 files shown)


src/containers/CheckoutPage/CheckoutPageWithPayment.js
```diff
@@ -506,6 +511,12 @@ export const CheckoutPageWithPayment = props => {
     orderData?.deliveryMethod === 'shipping' &&
     !hasTransactionPassedPendingPayment(existingTransaction, process);
 
+  const listingLocation = listing?.attributes?.publicData?.location;
+  const isBooking = processName === BOOKING_PROCESS_NAME;
+  const isPurchase = processName === PURCHASE_PROCESS_NAME;
+  const showPickUpLocation = isPurchase && orderData?.deliveryMethod === 'pickup';
+  const showLocation = isBooking && listingLocation?.address;
```
SariSaar, line 518: For symmetry, would it make sense to name this showBookingLocation, since it's specifically on bookings? Or is this something that's going to eventually be applied to negotiation location as well?

PR #681, opened 2025-10-28
Only set TransactionPage pageHeading if processName resolves

SariSaar:
With a full page load, TransactionPage will throw a missing translation error if pageHeading is set before the transaction process name is available. This fix only sets pageHeading after the transaction has loaded and process name can be used.


Gnito (approved):
LGTM!

PR #691, opened 2025-11-04
Accessibility improvements

Amanda-Kauppinen:
Improves template accessibility further

  - Improves contrast for various colors throughout the Template
  - Increases area around clickable elements
  - Updates various component roles

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2025-XX-XX
 
+- [add] Improve template accessibility further
+
+  - Improved contrast for various colors throughout the Template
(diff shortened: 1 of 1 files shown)


src/containers/CheckoutPage/CheckoutPage.module.css
```diff
@@ -243,7 +243,7 @@
   color: var(--colorFail);
 }
 .notFoundError {
-  color: var(--colorAttention);
+  color: var(--colorAttentionDark);
```
Gnito: This is strange. I'm not sure why error text is yellow in the first place. I.e. on purpose or is this some kind of copy-paste bug. 

No need to change this to red on this PR, but perhaps it's worth testing this error during some cooldown week.

PR #701, opened 2025-11-17
Fix pagination assignment to use correct path

shareoc:
Pagination links were not being rendered on the manage listings page. Fixed pagination state assignment to correctly access metadata from `action.payload.data.meta` instead of `action.payload.meta`


Gnito (approved):
(Run `yarn run format-docs`)

Otherwise: LGTM!

PR #729, opened 2025-12-18
Improve banned and deleted user handling on Transaction Page

SariSaar:
This change improves banned and deleted user handling on the transaction page:
- For inquiry and negotiation processes, the initial message is saved in the transaction's protected data instead of sent as a message. Now, when the sender is banned, those messages are replaced with the same default marketplace texts as regular messages.
- Transaction process state specific extra messages are hidden for banned and deleted users.
- Transaction page action buttons are disabled when the counterparty is banned or deleted, because transactions cannot move forward in that circumstance.

Screenshot 2025-12-18 at 10 33 00

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2025-XX-XX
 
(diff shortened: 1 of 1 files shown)


SariSaar:
TODO: Fix example files in /styleguide to use the new logic

PR #737, opened 2026-01-09
Add transaction fields based on Console configuration

SariSaar:
This PR prepares the codebase for using transaction field configurations that will be a part of the listing type asset. Operators can define transaction fields per listing type.

## Relevant changes

This PR introduces the following changes:
- Transaction fields
- CustomExtendedDataSection directory
- Inquiry message moved to protected data for booking and purchase processes

### Transaction fields

Transaction field content is collected in different places depending on the transaction process associated with the listing type.

With inquiry transactions, transaction fields inputs are shown on CheckoutPageWithInquiryProcess page where the customer sends the first message.
Screenshot 2026-02-06 at 11 56 20

With purchase and booking transactions, transaction field inputs are shown on CheckoutPageWithPayment page when the customer makes the payment.
Screenshot 2026-02-06 at 11 51 01

For these two processes, transaction field inputs are always only shown to the customer.

With the negotiation process, the operator can define whether the field is filled out by the customer or the provider. 

If the field is defined for the customer, the field inputs are shown on RequestQuotePage where the customer requests a quote from the provider in a regular negotiation flow.
Screenshot 2026-02-05 at 10 12 49

If the field is defined for the customer, the field inputs are shown on MakeOfferPage where the provider either makes or updates their offer to the customer.
Screenshot 2026-02-05 at 10 15 11

For all processes, the data stored with transaction fields is shown on TransactionPage.
Screenshot 2026-02-05 at 10 15 41

### CustomExtendedDataSection directory

In addition to this feature change, this PR also adds a CustomExtendedDataSection directory to `src/components` that contains shared display components for custom extended data fields. Previously, display components were duplicated in ProfilePage and ListingPage directories, and this PR changes that logic so that all contexts displaying custom extended data fields – ProfilePage, ListingPage, TransactionPage – now use shared components.

### Inquiry message moved to protected data for booking and purchase processes

Up until now, there have been two types of handling for the initial message in a transaction. In booking and purchase processes, the initial message from CheckoutPageWithPayment was sent with sdk.messages.send, and in inquiry and negotiation processes, the message was saved in the transaction's protected data. This PR updates booking and purchase processes to also save the initial message in the transaction's protected data for unified handling.

## Translation changes

```diff
+ "TransactionPage.detailsTitle": "Details",
+ "TransactionPanel.defaultMessageHeading": "Additional information",
```

### DIFF
--- a/src/components/CustomExtendedDataSection/README.md
+++ b/src/components/CustomExtendedDataSection/README.md
@@ -1,0 +1,49 @@
+# CustomExtendedDataSection
+
+This is a reusable component for displaying custom extended data fields configured via Sharetribe
+hosted assets. This component is used across multiple entity types (listings, user profiles, and
+transactions) to render custom fields in a consistent way.
+
+## Overview
+
+`CustomExtendedDataSection` is a generic component that displays extended data based on custom field
+configurations. The component renders different field types using specialized subcomponents based on
+the field's `schemaType`.
+
+## Supported Schema Types
+
+The component handles six different schema types:
+
+| Schema Type       | Display Component     | Description                                                |
+| ----------------- | --------------------- | ---------------------------------------------------------- |
+| `enum`            | `SectionDetails`      | Single-select dropdown values displayed as key-value pairs |
+| `long`            | `SectionDetails`      | Numeric values displayed as key-value pairs                |
+| `boolean`         | `SectionDetails`      | Yes/No values displayed as key-value pairs                 |
+| `multi-enum`      | `SectionMultiEnum`    | Multiple selection fields displayed with PropertyGroup     |
+| `text`            | `SectionText`         | Long-form text                                             |
+| `youtubeVideoUrl` | `SectionYoutubeVideo` | Embedded YouTube videos                                    |
+
+## Extended Data Scopes
+
+The component currently supports three extended data scopes:
+
+- **`public`** - Publicly visible data (listings, user profiles)
+- **`protected`** - Transaction protected data (only visible to transaction participants)
+- **`metadata`** - Internal metadata (less commonly displayed)
+
+The scope is determined by the field configuration's `scope` property. The `pickCustomFieldProps()`
+helper automatically extracts values from the correct extended data scope.
+
+## Entity Type Filtering
+
+For listings and users, fields can be restricted to specific entity types (e.g., only show certain
+fields for "product" listings vs "service" listings). The component respects these restrictions when
+`entityTypeKey` is provided.
+
+For transactions, pass `null` as `entityTypeKey` since transaction fields don't use entity type
+filtering. They use role-based filtering instead (customer vs provider).
+
+## Related Files
+
+- `src/util/fieldHelpers.js` - Helper functions for preparing field data
+- `src/components/CustomExtendedDataField/` - Form inputs for the corresponding configurations


src/containers/TransactionPage/TransactionPanel/TransactionPanel.js
```diff
@@ -343,7 +350,15 @@ export class TransactionPanelComponent extends Component {
...
-
+            {showCustomTransactionFields ? (
+              <CustomTransactionFields
+                protectedData={protectedData}
+                transactionFieldConfigs={transactionFields}
+                intl={intl}
+                className={css.customFieldsContainer}
+                role={'customer'}
+              />
+            ) : null}
```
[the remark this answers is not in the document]
Gnito: Note: for some strange reason, order details container swaps the Delivery info & booking location above the mobile price breakdown. (We'd need a "code archeologist" to understand why it's using "order" css property.)

So, the actual order is therefore 
- transaction fields
- other order details (delivery & booking location)
- mobile price breakdown

=> The only change I'm looking for here is that all the transaction fields are set above the `.orderDetails` div.

src/containers/ListingPage/CustomListingFields.js
```diff
@@ -35,27 +33,52 @@ const CustomListingFields = props => {
...
-          <SectionTextMaybe key={key} {...fieldProps} />
-        ) : schemaType === SCHEMA_TYPE_YOUTUBE ? (
-          <SectionYoutubeVideoMaybe key={key} {...fieldProps} />
-        ) : null;
-      })}
-    </>
+    <CustomExtendedDataSection
+      sectionDetailsProps={sectionDetailsProps}
+      propsForCustomFields={propsForCustomFields}
+      page="ListingPage"
```
Gnito: Is there a possibility in the future that one page could contain multiple <CustomExtendedDataSection> components? Perhaps rendering publicData and privateData keys separately on some listing or user related pages?

Because of that, I'd suggest changing this property to something more id or context related: e.g. `idPrefix` is something that's been used before.
Then it would make it more reasonable to write something like `idPrefix="listingPrivateData"`

src/containers/CheckoutPage/CheckoutPageTransactionHelpers.js
```diff
@@ -303,19 +302,12 @@ export const processCheckoutWithPayment = (orderParams, extraPaymentParams) => {
     return orderPromise;
   };
 
-  //////////////////////////////////
-  // Step 4: send initial message //
-  //////////////////////////////////
-  const fnSendMessage = fnParams => {
-    const orderId = fnParams?.id;
-    return onSendMessage({ id: orderId, message });
```
Gnito: Check if it's possible to remove the whole thunk from CheckoutPage.duck.js

src/containers/RequestQuotePage/RequestQuoteForm/RequestQuoteForm.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Form as FinalForm } from 'react-final-form';
+import arrayMutators from 'final-form-arrays';
```
Gnito: Check my comment on StripePaymentForm.js

src/containers/MakeOfferPage/MakeOfferPage.js
```diff
@@ -42,6 +42,7 @@ import MakeOfferForm from './MakeOfferForm/MakeOfferForm.js';
 import { makeOffer } from './MakeOfferPage.duck.js';
 
 import css from './MakeOfferPage.module.css';
+import { pickTransactionFieldsData } from '../../util/fieldHelpers.js';
```
Gnito: Move this next to other util imports

src/util/fieldHelpers.js
```diff
@@ -227,3 +232,114 @@ export const isValidCurrencyForTransactionProcess = (
...
+      return isKnownSchemaType && isTransactionScope && showField
+        ? [
+            ...pickedFields,
+            {
+              key: namespacedKey,
+              name: namespacedKey,
+              fieldConfig: fieldConfig,
+              defaultRequiredMessage: intl.formatMessage({
+                id: 'CustomExtendedDataField.required',
+              }),
```
Gnito: This seems to break our convention of translation key pointing to the component that renders it. (...and I now realise that we have already done this with _getPropsForCustomUserFieldInputs_)

Screenshot 2026-01-20 at 19 37 10

A translation key syntax has given a hint where the relevant code is. I.e. a key like 'CustomExtendedDataField.required' should be found from a component called CustomExtendedDataField.

So, the "defaultRequiredMessage" property could be included here:
https://github.com/sharetribe/web-template/blob/main/src/components/CustomExtendedDataField/CustomExtendedDataField.js#L230

Perhaps, it would still make sense to keep util files clean of intl.formatMessage calls:
Screenshot 2026-01-20 at 19 47 25

src/containers/CheckoutPage/CheckoutPageWithInquiryProcess.js
```diff
@@ -33,6 +35,10 @@ import DetailsSideCard from './DetailsSideCard';
 import MobileListingImage from './MobileListingImage';
 
 import css from './CheckoutPage.module.css';
+import {
+  getPropsForCustomTransactionFieldInputs,
+  pickTransactionFieldsData,
+} from '../../util/fieldHelpers.js';
```
Gnito: Move these next to other util imports

src/containers/CheckoutPage/StripePaymentForm/StripePaymentForm.js
```diff
@@ -20,11 +21,13 @@ import {
   IconSpinner,
   SavedCardDetails,
   StripePaymentAddress,
+  CustomExtendedDataField,
 } from '../../../components';
 
 import ShippingDetails from '../ShippingDetails/ShippingDetails';
 
 import css from './StripePaymentForm.module.css';
+import { getPropsForCustomTransactionFieldInputs } from '../../../util/fieldHelpers';
```
Gnito: Move this next to other util imports

src/containers/MakeOfferPage/MakeOfferForm/MakeOfferForm.js
```diff
@@ -21,6 +23,7 @@ import {
 } from '../../../components/index.js';
 
 import css from './MakeOfferForm.module.css';
+import { getPropsForCustomTransactionFieldInputs } from '../../../util/fieldHelpers.js';
```
Gnito: Move this next to other util imports

src/containers/CheckoutPage/StripePaymentForm/StripePaymentForm.js
```diff
@@ -5,6 +5,7 @@
  */
 import React, { Component } from 'react';
 import { Form as FinalForm } from 'react-final-form';
+import arrayMutators from 'final-form-arrays';
```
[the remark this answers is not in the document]
SariSaar: Including a multi-enum field throws an error on the Checkout page for missing arrayMutators if this is not included. I'm assuming it's because it uses FieldCheckboxGroup which in turn uses FieldArray, so even though the fields aren't added dynamically, FieldArray depends on arrayMutators.

PR #738, opened 2026-01-09
Update marketplace text

thomasmalbaux:
When the default-negotiation process was added (https://github.com/sharetribe/web-template/pull/648), some Marketplace texts were updated to fit the new process. One was missed, and it is now updated in this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,15 @@ way to update this template, but currently, we follow a pattern:
 
 ---
 
-## Upcoming version 2025-XX-XX
+## Upcoming version 2026-XX-XX
 
 - [add] Update a Marketplace text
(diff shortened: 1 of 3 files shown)


Gnito:
@thomasmalbaux I don't think this is correct. Those tabs are about roles (provider & curstomer aka who receives payment and who pays) - and listing author is not a role as nowadays customer can be author in addition to provider.

Screenshot 2026-01-12 at 13 52 36

Where did you encounter this issue? (And do you have user types and existing user entities set correctly?)

thomasmalbaux:
@Gnito You're exactly right. I noticed the differences between the two strings directly in the json translations file, and I couldn't exactly remember which one was right and I got confused.

I've now updated the correct value.

Does it look OK to you, now? I just want your confirmation to make sure I got it right this time. Thanks!

PR #759, opened 2026-02-02
WIP don't Add AGENTS.md

SariSaar:
~Add an AGENTS.md file to add guidance to coding agents.~

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,335 @@
+# AGENTS.md - AI Agent Code Generation Guide
+
+Sharetribe Web Template: React 18 + Redux Toolkit + Final Form + Express.js + Server-Side Rendering
+
+## Project Overview
+
+- **Type**: React-based marketplace web application with Server-Side Rendering
+- **Tech Stack**: React 18, Redux Toolkit, Final Form, Express.js, Sharetribe SDK
+- **Purpose**: Full-featured two-sided marketplace template with multiple transaction processes
+
+## Critical Conventions
+
+### 1. File Structure
+
+**Component:**
+
+```
+/ComponentName/
+  ComponentName.js              # Main component
+  ComponentName.module.css      # CSS Module (required)
+  ComponentName.example.js      # Styleguide example (optional, for reusable components)
+  ComponentName.test.js         # Tests (optional)
+```
+
+**Page:**
+
+```
+/PageName/
+  PageName.js                   # Main page
+  PageName.module.css           # Styles
+  PageName.duck.js              # Redux state (Redux Toolkit)
+  PageName.test.js              # Tests
+```
+
+### 2. Component Template
+
+**Always use this template when creating new components:**
+
+```javascript
+import React from 'react';
+import classNames from 'classnames';
+import { FormattedMessage } from 'react-intl';
+import { IconComponent } from '../../components';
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,335 @@
...
+`export { default as ComponentName } from './ComponentName/ComponentName';`
+
+Pages in `/src/routing/routeConfiguration.js`:
+
+```javascript
+const PageName = loadable(() =>
+  import(/* webpackChunkName: "PageName" */ '../containers/PageName/PageName')
+);
+// Route: { path: '/page', name: 'PageName', component: PageName, auth: true }
+```
```
Gnito: I'd omit this one. 
- The `src/components/index.js` is relevant tip when creating a component or working with directory structure. (headings 1 & 2)
- The latter is repeated zillion times on routeConfiguration.js file - no need to repeat it here.

AGENTS.md
```diff
@@ -0,0 +1,335 @@
...
+
+### 10. API Calls
```
Gnito: There are several topics that are about API.

1. Most common: make Marketplace API calls on src/containers/<PageName>.duck.js file.
  - Don't add data loading calls to useEffect hook
  - Instead, add page-level <PageName>.duck.js file which contains loadData function. Register the loadData on src/containers/pageDataLoadingAPI.js and routeConfiguration.js file.

AGENTS.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Don't:
+
+❌ Class components, inline styles, hard-coded text, manual Redux, wrong import order
+
+## Checklist
+
+- [ ] Functional components | Correct import order | CSS Modules | `rootClassName`/`className` props
+- [ ] All text i18n | Components exported in `/src/components/index.js` | Pages lazy-loaded
+- [ ] Redux Toolkit | Final Form | Mobile-first | PropTypes
```
Gnito: Do we need this?

AGENTS.md
```diff
@@ -0,0 +1,335 @@
...
+**Testing:** Use `renderWithProviders()`, `fakeIntl`, `createUser()`, `createListing()` from test
+helpers. Co-locate tests with components.
+
+**Config:** `useConfiguration()`, `useRouteConfiguration()` from context
+
+### 12. Hosted Configuration Assets
+
+**DO NOT modify hosted asset configurations directly in code.** The marketplace configuration is
+managed through **Sharetribe Console** and fetched at runtime via the Asset Delivery API.
+
```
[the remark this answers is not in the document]
SariSaar: > could be good to include a practical example of how to access specific asset values, e.g. fetch the branding color or listing types via the SDK

There's an example under "Configuration Access in Components" on how to get `marketplaceColor` and `listingTypes` with the useConfiguration() hook.

AGENTS.md
```diff
@@ -0,0 +1,335 @@
...
+  const marketplaceColor = branding.marketplaceColor;
+  const listingTypes = listing.listingTypes;
+
+  // ...
+};
+```
+
+#### Important Rules
+
+1. **Never hardcode** configuration values that can be set in Console (colors, listing types, user
```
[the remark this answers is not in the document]
SariSaar: Good point, I'll reformulate!

PR #760, opened 2026-02-03
Add support for hosted sort configurations

Amanda-Kauppinen:
Adds support for merging hosted listing search sort configurations (from Console) with default configurations defined in Template

image

### DIFF
--- a/src/util/configHelpers.js
+++ b/src/util/configHelpers.js
@@ -1445,7 +1445,7 @@ const validSortConfig = config => {
   return { active, queryParamName, relevanceKey, relevanceFilter, conflictingFilters, options };
 };
 
-const mergeSortConfig = (hostedSortConfig, defaultSortConfig) => {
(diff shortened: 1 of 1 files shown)


Gnito (approved):
LGTM!
(Update the description of the PR.)

src/containers/SearchPage/SearchPage.duck.js
```diff
@@ -237,6 +237,27 @@ const searchListingsPayloadCreator = ({ searchParams, config }, thunkAPI) => {
...
+      return { sort: sortParam };
+    }
+
+    // No sort parameter needed when keyword search is used or sort config is inactive
+    if (relevanceEnabled && (hasKeywords || !sortConfig.active)) {
+      return {};
+    }
+
+    // Fall back to default sort
+    return { sort: defaultSort };
```
Gnito: If we have multiple sort options or just 1, this works as intended. (At least, I think so.)
However, when there is no sort options given, we should omit the sort parameter altogether. 
(or use createdAt, which is likely the default that API uses as default.)

PR #765, opened 2026-02-05
Add AGENTS.md file

SariSaar:
Add an AGENTS.md file to provide context for LLMs on codebase conventions, cross-system dependencies, and best practices in the Sharetribe Web Template.

NOTE: When applying this PR to your custom developed application, be sure to review the contents of this file and make updates if you have adopted workflows and patterns that diverge from the ones described here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2026-XX-XX
 
+- [add] Add AGENTS.md file to provide context for LLMs.
+  [#765](https://github.com/sharetribe/web-template/pull/765)
 - [fix] Routes.js: fix a bug with hash selector. (CSS.escape was not used correctly.)
   [#774](https://github.com/sharetribe/web-template/pull/774)
 - [fix] SearchPage.shared.js: fix a bug with keywords search type. (SortBy doesn't pick relevance


AGENTS.md
```diff
@@ -0,0 +1,168 @@
...
+  - Page-level .duck.js files are used from a specific page, and they live in the corresponding page
+    directory.
+
+## API calls
+
+- Use the Sharetribe SDK to call Sharetribe Marketplace API
+- For other API calls, including to the app's own server, create a function in `src/util/api.js` and
+  follow the established conventions.
+- Make API calls for a page (GET, POST, PUT, DELETE) in a `src/containers/{PageName}.duck.js` file.
+  See `src/containers/ListingPage/ListingPage.duck.js` for an example.
```
Gnito: Perhaps something like:

Make API calls for a page (using SDK) in a `src/containers/{PageName}.duck.js` file.

AGENTS.md
```diff
@@ -0,0 +1,168 @@
...
+- Do not use a `useEffect` hook to make API data loading calls. Instead, add a page-level .duck.js
+  file that contains a loadData function. Register the loadData in
+  `src/containers/pageDataLoadingAPI.js` and `routeConfiguration.js` files following the existing
+  conventions in those files.
```
Gnito: This is our pattern for data flow and it's there to help SSR => data loading needs for a route are found through a specific function. (so, I like that this is written here explicitly.)

However, it would be good to mention in the PRs description that customizers might want to change the content. And this pattern of using loadData function is not be relevant for components that load data only on client-side. (We don't have many components that are purely for client-side rendering on Template - map-related components, of course, work like that, and the upcoming "featured listings" section for Pages.)

Gnito (approved):
LGTM

PR #779, opened 2026-02-19
Custom metadata fields

shareoc:
Adds support for rendering custom metadata fields on listing and user pages, and using them as search filters.

Custom metadata fields in listings can be defined like regular listing fields, but must be scoped as metadata fields. Custom metadata fields also support a displayOnListingPage flag which can be set to false to not render the field on the listing page. Like regular listing fields, custom metadata fields can be configured as search filters on the search page.

```json
"scope": "metadata",
"displayOnListingPage": false
```

Defining a user field as a metadata field does not include the option to hide the field from the user page, and unlike private and public user fields, lacks the option to add the field to the sign-up form.

### DIFF
--- a/src/containers/SearchPage/SearchPageWithMap.js
+++ b/src/containers/SearchPage/SearchPageWithMap.js
@@ -15,6 +15,7 @@ import {
   isMainSearchTypeKeywords,
   isOriginInUse,
   getQueryParamNames,
+  constructQueryParamName,
 } from '../../util/search';
 import {
   NO_ACCESS_PAGE_USER_PENDING_APPROVAL,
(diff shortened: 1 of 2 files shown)


src/util/search.js
```diff
@@ -22,7 +33,8 @@ export const parseSelectFilterOptions = uriComponentValue => {
  * @param {String} scope Scope extracted from listingExtendData config.
  */
 export const constructQueryParamName = (key, scope) => {
-  const prefixedKey = scope === 'meta' ? `meta_${key}` : `pub_${key}`;
+  // we check both meta and metadata to ensure backwards compatibility (assets use metadata)
```
Gnito: Update also configListing.js: There's scope mentioned there, but only public vs private. We should add metadata there. (I.e. if customizers have made changes to allow metadata they might have used "meta" before and if someone new ends up to that file, it would be good that it reflects the current options.)

PR #783, opened 2026-02-23
Remove location heading for listing types with no location

SariSaar:
Transaction page for booking listings showed a "Location" heading after the booking was accepted, even if the listing type specified `defaultFields.location` as `false`.

Screenshot 2026-02-23 at 17 02 02


Gnito (approved):
LGTM

PR #792, opened 2026-02-25
Eject from CRA

Gnito:
## Webpack configs in the repo

Having these configs in the repo gives more power for customizers - but that also means that it is more likely that the configs end up causing issues that's really difficult to fix. Therefore, unless you know what you are doing, you should avoid modifying them.

## Background

Sharetribe Web Template uses a fork of [Create React App](https://github.com/facebook/create-react-app) (CRA), called _sharetribe-scripts_. The fork was there because Template supports also server-side rendering, code-splitting, etc.

Since Create React App has been deprecated and has not received updates for a long time, it has become more of a blocker for updating the dependency tree through a several interconnected packages that the CRA consisted of. The sharetribe-scripts is just fork of react-scripts package.

## Ejection

Ejection extracted build,start, and test scripts to scripts directory. In addition, it created a config directory to the root. The config then contained Webpack configs and related files. It also referenced other CRA-related packages, which also needs to be inlined (or forked). This PR inlined those packages (_babel-preset-react-app_, _eslint-config-react-app_, and _react-dev-utils_) to the _**config/**_ directory. 

## Extra notes

The private [postcss-apply](https://github.com/sharetribe/create-react-app/tree/master/packages/react-scripts/postcss/postcss-apply) plugin is not included anymore. It was a fallback setup for a deprecation on a past version of the sharetribe-scripts.

Currently, this repository does not use Eslint. The related devDependencies are still included here, but the full eslint-ruleset was reduced when updating those packages. You could expand this rough config to your own needs. This PR tested the included setup a bit and changed some very old eslint code comments because of it. (When Prettier was released, we dropped Eslint support - but there are certainly arguments for taking modern Eslint back into the tool box. However, we didn't do that in this PR.)

CRA also contains support for other features that are not used in this project. E.g. TypeScript.
Most devDependencies related to these features are included in some form. 
(Note: if you want to use SASS or Tailwind, you need to figure out the best way to do that. [It was not included.](https://github.com/facebook/create-react-app/blob/main/packages/react-scripts/config/webpack.config.js#L161-L173))

JavaScript world is also slowly moving from CJS modules to ES modules. This PR tries to update all the dependencies as far as possible, but because some deps have moved to ES-only setup, and some are not, not all the dependencies are yet on the most recent versions. This is something we need to tackle in the future.

The `dotenv` package was updated quite much - and that included changes to how hashtags and backticks are used. So, pay attention to environment variables working when testing this PR.

The `babel-jest` dependency does not support [NodeJS@v23](https://github.com/jestjs/jest/pull/15640) at all. So, we need to drop that support too.

Jest update updated jsdom library, which didn't keep 'this' instance on server/api-util/lineItemsHelpers.js when other functions on module were referenced. Changed to use 'exports'.

Some SASS-related loaders and configs were not included.

Most of the 12k of additions/deletions on this PR is coming from _yarn.lock_ file.

babel-plugin-istanbul causes unnecessary warnings on installation
"Workspaces can only be enabled in private projects."


config/webpack.config.js
```diff
@@ -0,0 +1,723 @@
...
+      rules: [
+        // Handle node_modules packages that contain sourcemaps
+        shouldUseSourceMap && {
+          enforce: 'pre',
+          exclude: /@babel(?:\/|\\{1,2})runtime/,
+          test: /\.(js|mjs|jsx|ts|tsx|css)$/,
+          loader: require.resolve('source-map-loader'),
+        },
+        {
+          // "oneOf" will traverse all following loaders until one will
```
SariSaar: Is this something where we'd need to add a comment about the sharetribeWebpackConfig.js checkConfigStructure breaking if this array is changed, or is that handled somewhere else in the code?

PR #812, opened 2026-03-24
Add short text field

Amanda-Kauppinen:
This PR adds support for a new user, listing, and transaction field schema type: `shortText`

### DIFF
--- a/src/components/CustomExtendedDataField/CustomExtendedDataField.js
+++ b/src/components/CustomExtendedDataField/CustomExtendedDataField.js
@@ -5,6 +5,7 @@ import { useIntl } from '../../util/reactIntl';
 import {
   SCHEMA_TYPE_ENUM,
   SCHEMA_TYPE_MULTI_ENUM,
+  SCHEMA_TYPE_SHORT_TEXT,
   SCHEMA_TYPE_TEXT,
   SCHEMA_TYPE_LONG,
   SCHEMA_TYPE_BOOLEAN,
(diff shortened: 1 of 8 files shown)


src/components/CustomExtendedDataField/CustomExtendedDataField.js
```diff
@@ -85,6 +86,32 @@ const CustomFieldMultiEnum = props => {
   ) : null;
 };
 
+// TODO: replace placeholder text and update styling as needed -- or perhaps merge this with CustomFieldText?
```
Gnito: I guess, this is outdated comment.
(It's probably better to have each of these inputs handled separately per schema type => no need to merge this to CustomFieldText at this point.)

src/components/CustomExtendedDataSection/SectionDetails.js
```diff
@@ -5,6 +5,8 @@ import { FormattedMessage } from '../../util/reactIntl';
 
 import { Heading } from '../../components';
 
+import { richText } from '../../util/richText';
```
Gnito, line 6: This is imported in a wrong place (utils after library imports and before component imports). Otherwise, there is somewhat high chance that we/customizers end up with circular dependency issues.

src/components/CustomExtendedDataSection/SectionDetails.js
```diff
@@ -26,12 +28,17 @@ const SectionDetails = props => {
...
-        {existingFields.map(detail => (
-          <li key={detail.key} className={css.detailsRow}>
-            <span className={css.detailLabel}>{detail.label}</span>
-            <span>{detail.value}</span>
-          </li>
-        ))}
+        {existingFields.map(detail => {
+          const valueContent = richText(detail.value, {
+            linkify: true,
+          });
```
Gnito, line 37: This is now a bit wider setup than needed.
I think it's good for now - but if we ever want to add for example "Custom link field", this becomes cumbersome catch all (as link field would need separate value handling for external link vs in-app link etc. and it can't use richText function at that point).

Therefore, it would be good to add code-comment here. Something like this:


```suggestion
          // Note: currently, we are auto-linking any value of these key-value details.
          // If this behaviour needs to be changed later (e.g. due to separate "custom link field"), 
          // we need to add extra flag (linkify: true) through getDetailCustomFieldValue function in util/fieldHelpers.js
          const valueContent = richText(detail.value, {
            linkify: true,
          });
```

Gnito: (Or the richText could be applied already inside fieldHelpers.js > getDetailCustomFieldValue - but let's leave that decision for the future.)

src/components/CustomExtendedDataSection/SectionDetails.js
Amanda-Kauppinen: For some reason this UI is not updating for me, but this is on line 6 now, and components import on line 8

src/components/CustomExtendedDataSection/SectionDetails.js
Amanda-Kauppinen: comment added

Gnito (approved):
LGTM

PR #816, opened 2026-03-31
Fix return value in CheckoutPage pay-and-save-card flow

SariSaar:
The Checkout page flow for saving a payment method after purchase was throwing an error instead of redirecting. Fix the value in the return statement to correctly return the order id.


walterrw (approved):
LGTM! Left a comment about whether to include or leave out other params, but couldn't find any places where other params would be expected.

PR #822, opened 2026-04-09
stripe fixes based on new stripe rules for europe

rakeshIcodelabs:
https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025

### DIFF
--- a/src/util/stripeConnect.js
+++ b/src/util/stripeConnect.js
@@ -1,0 +1,63 @@
+const countriesRequiringSoleProprietorshipForIndividuals = new Set([
+  'AT',
+  'BE',
+  'BG',
+  'CH',
+  'CY',
+  'CZ',
+  'DE',
+  'DK',
+  'EE',
+  'ES',
+  'FI',
+  'FR',
+  'GB',
+  'GI',
+  'GR',
+  'HR',
+  'HU',
+  'IE',
+  'IS',
+  'IT',
+  'LI',
+  'LT',
+  'LU',
+  'LV',
+  'MT',
+  'NL',
+  'NO',
+  'PL',
+  'PT',
+  'RO',
+  'SE',
+  'SI',
+  'SK',
+]);
+
+export const requiresSoleProprietorshipAccount = country =>
+  countriesRequiringSoleProprietorshipForIndividuals.has(country);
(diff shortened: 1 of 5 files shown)


src/util/stripeConnect.js
```diff
@@ -0,0 +1,63 @@
...
+  'LV',
+  'MT',
+  'NL',
+  'NO',
+  'PL',
+  'PT',
+  'RO',
+  'SE',
+  'SI',
+  'SK',
```
Gnito, line 34: @rakeshIcodelabs thank you for the pull request! I think it needs to be adapted a bit, but it highlights one issue that we have overlooked.

So, my current understanding is that only Netherlands is affected by this requirement about 'individual' business type - not all of these listed countries.

First of all, the article seems to say that:

> If you choose not to migrate to Stripe-hosted or Embedded onboarding, then you need to address the following updates:
> 
> - [Know Your Customer (KYC) verification](https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025#know-your-customer-(kyc)-verification)
> - [Ultimate beneficial owner (UBO) and director relationship verification](https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025#ubo-director-verification)
> - [Netherlands business registration (KvK) requirements](https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025#netherlands-business-registration-requirements)
> - [New error codes](https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025#new-error-codes)
> 
https://docs.stripe.com/connect/upcoming-requirements-updates?program=eu-2025#api-integration-overview

=> Template uses Stripe's hosted onboarding, so these should be mostly covered.

That being said, [Stripe's API](https://docs.stripe.com/api/accounts/object) seems to say that business_type can be nullable. So, perhaps we cause issues (by enforcing business_type when creating a connect account) - even if we use Stripe's hosted onboarding.

So, my plan is to create another version of this, which only modifies Netherlands business_type requirement. (I'll pick this commit there as a starting point.)

Gnito:
Closing in favour of:
https://github.com/sharetribe/web-template/pull/824

PR #826, opened 2026-04-13
Referral ID

Amanda-Kauppinen:
Adds template support for the upcoming referral link feature.

When a user follows a referral link, the referral ID is validated against user referrals defined in Console and checked against a matching user type. On sign-up, the referral ID is saved to the user's private data.

Local storage is used to persist the referral ID, with a 90-day expiry. Referral data is captured on any page load, so any page can serve as an entry point for a referral link.

Note for custom code: If URL parameters are passed directly to API queries anywhere in your custom code, referral parameters could be unintentionally included. Make sure to filter them out before making API calls. See [SearchPage.duck.js](https://github.com/sharetribe/web-template/blob/cb2d59c91d101c33b276368315be8ff2b4519359/src/containers/SearchPage/SearchPage.duck.js#L282-L291) for an example of the filtering.

### DIFF
--- a/src/containers/AuthenticationPage/AuthenticationPage.helpers.js
+++ b/src/containers/AuthenticationPage/AuthenticationPage.helpers.js
@@ -2,6 +2,14 @@ import Cookies from 'js-cookie';
 
 import { isEmpty } from '../../util/common';
 import { pickUserFieldsData, addScopePrefix } from '../../util/userHelpers';
+import { pickReferralData } from '../../util/sessionStorageHelpers';
+
+// Returns full userType config based on selected userType
+const getUserTypeConfig = (userType, userTypes) => {
+  return userTypes.find(config => {
+    return config.userType === userType;
+  });
+};
 
 /**
  * Filters out configured user-field entries, returning only the remaining key/value pairs.
@@ -39,17 +47,22 @@ export const getNonUserFieldParams = (values, userFieldConfigs) => {
  * @param {Array} userFields - User field configurations
  * @returns {{ publicData: Object, privateData: Object, protectedData: Object } | {}}
  */
-export const getExtendedDataMaybe = (submitValues, userType, userFields) => {
+export const getExtendedDataMaybe = (submitValues, userType, userFields, extraData) => {
+  const { publicData, privateData, protectedData } = extraData;
+
   return !isEmpty(submitValues)
     ? {
         publicData: {
+          ...publicData,
           userType,
           ...pickUserFieldsData(submitValues, 'public', userType, userFields),
         },
         privateData: {
+          ...privateData,
           ...pickUserFieldsData(submitValues, 'private', userType, userFields),
         },
         protectedData: {
+          ...protectedData,
           ...pickUserFieldsData(submitValues, 'protected', userType, userFields),
           // If the form has any additional values, pass them forward as user's protected data
           ...getNonUserFieldParams(submitValues, userFields),
@@ -67,9 +80,13 @@ export const getExtendedDataMaybe = (submitValues, userType, userFields) => {
  * @param {Array} params.userFields
  * @returns {(values: Object) => void}
  */
-export const getHandleSubmitSignup = ({ submitSignup, userFields }) => values => {
+export const getHandleSubmitSignup = ({ submitSignup, userFields, userTypes }) => values => {
   const { userType, email, password, fname, lname, displayName, ...rest } = values;
   const displayNameMaybe = displayName ? { displayName: displayName.trim() } : {};
+
+  // Set referral to user private data if it exists and is valid
+  const userTypeConfig = getUserTypeConfig(userType, userTypes);
+  const extraPrivateData = pickReferralData(userTypeConfig);
 
   const submitParams = {
     email,
@@ -77,7 +94,9 @@ export const getHandleSubmitSignup = ({ submitSignup, userFields }) => values =>
     firstName: fname.trim(),
     lastName: lname.trim(),
     ...displayNameMaybe,
-    ...getExtendedDataMaybe(rest, userType, userFields),
+    ...getExtendedDataMaybe(rest, userType, userFields, {
+      privateData: extraPrivateData,
+    }),
   };
 
   submitSignup(submitParams);
@@ -93,7 +112,12 @@ export const getHandleSubmitSignup = ({ submitSignup, userFields }) => values =>
  * @param {Array} params.userFields
  * @returns {(values: Object) => void}
  */
-export const getHandleSubmitConfirm = ({ authInfo, submitSingupWithIdp, userFields }) => values => {
+export const getHandleSubmitConfirm = ({
+  authInfo,
+  submitSingupWithIdp,
+  userFields,
+  userTypes,
+}) => values => {
   const { idpToken, email, firstName, lastName, idpId } = authInfo;
 
   const {
@@ -115,8 +139,14 @@ export const getHandleSubmitConfirm = ({ authInfo, submitSingupWithIdp, userFiel
     ...(newLastName !== lastName && { lastName: newLastName }),
   };
 
+  // Set referral to user private data if it exists and is valid
+  const userTypeConfig = getUserTypeConfig(userType, userTypes);
+  const extraPrivateData = pickReferralData(userTypeConfig);
+
   // Pass other values as extended data according to user field configuration
-  const extendedDataMaybe = getExtendedDataMaybe(rest, userType, userFields, true);
+  const extendedDataMaybe = getExtendedDataMaybe(rest, userType, userFields, {
+    privateData: extraPrivateData,
+  });
 
   submitSingupWithIdp({
     idpToken,
--- a/src/containers/AuthenticationPage/AuthenticationPage.helpers.js
+++ b/src/containers/AuthenticationPage/AuthenticationPage.helpers.js
@@ -2,7 +2,7 @@ import Cookies from 'js-cookie';
 
 import { isEmpty } from '../../util/common';
 import { pickUserFieldsData, addScopePrefix } from '../../util/userHelpers';
-import { pickReferralData } from '../../util/sessionStorageHelpers';
+import { pickReferralData } from '../../util/webStorageHelpers';
 
 // Returns full userType config based on selected userType
 const getUserTypeConfig = (userType, userTypes) => {
--- a/src/util/sessionStorageHelpers.js
+++ b/src/util/sessionStorageHelpers.js
@@ -5,10 +5,24 @@ import { isAfterDate } from '../util/dates';
 export const REFERRAL_ID_SESSION_STORAGE_KEY = 'referralSource';
 const HOUR_IN_MS = 60 * 60 * 1000;
 
+/**
+ * Returns true if the referral data's expiry timestamp has passed.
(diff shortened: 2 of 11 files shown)


src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,134 @@
+import { isEmpty } from './common';
+
+import { types as sdkTypes } from './sdkLoader';
+
+import { isAfterDate } from './dates';
```
Gnito: Change these to include the `util/`. That way we can more easily search for imports of "util/dates" as plain "dates" might be too vague search string.


```suggestion
import { isEmpty } from '../util/common';
import { types as sdkTypes } from '../util/sdkLoader';
import { isAfterDate } from '../util/dates';
```

src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,134 @@
...
+
+export const REFERRAL_ID_SESSION_STORAGE_KEY = 'referralSource';
+
+// TODO: add comments
+
+export const hasValidReferralLinkMaybe = (urlReferral, userTypes) => {
```
Gnito: We have used the *Maybe pattern in component files for subcomponents that might render something or not. I don't think it makes much sense to have that naming pattern here (and perhaps we should consider not using it even in component files).

```suggestion
export const hasValidReferralLink = (urlReferral, userTypes) => {
```

Gnito: What is that urlReferral? What kind of data? Is it full URL or object containing search params?

Perhaps, something like this would make more sense:
```suggestion
export const hasValidReferralParams = (urlReferralParams, userTypes) => {
```

src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,134 @@
...
+    return true;
+  }
+
+  return false;
+};
+
+export const storeReferralDataToSession = referralData => {
+  // Don't store the referral ID if it has already been stored
+  if (!hasReferralDataBeenStored(referralData)) {
+    if (window && window.sessionStorage && referralData) {
```
Gnito: Conditional chaining is nowadays possible. So, I'd change this a bit. 

```suggestion
    if (window?.sessionStorage && referralData) {
```

But perhaps this would be even more clear for customizers to read/understand:
```suggestion
    const isBrowser = typeof window !== 'undefined';
    if (isBrowser && window?.sessionStorage && referralData) {
```

src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,134 @@
...
+      };
+
+      const storableData = JSON.stringify(data, replacer);
+      window.sessionStorage.setItem(REFERRAL_ID_SESSION_STORAGE_KEY, storableData);
+    }
+  }
+};
+
+export const getStoredReferralDataFromSession = () => {
+  if (window && window.sessionStorage) {
```
Gnito:
```suggestion
  const isBrowser = typeof window !== 'undefined';
  if (isBrowser && window?.sessionStorage) {
```

src/util/includeScripts.js
```diff
@@ -180,6 +185,16 @@ export const IncludeScripts = props => {
...
 
+  // Check if session storage contains referral data
+  if (window && window.sessionStorage) {
+    const referralData = getStoredReferralDataFromSession();
+
+    // Referral data has expired, clear it from session storage
+    if (referralData != null && hasReferralDataExpired(referralData)) {
+      clearStoredReferralDataInSession();
+    }
+  }
```
Gnito: I think this should be turned into a separate helper (to be imported from util/sessionStorageHelpers.js).

Then perhaps on src/index.js, you could just import it

`import { clearExpiredReferralData } from './util/sessionStorageHelpers';`

And call it on line 147:
`clearExpiredReferralData();`

Gnito: After that, you could check if these even need to be exported from that util file - or could they just be helpers inside the sessionStorageHelpers.js file for the "clearExpiredReferralData" function.

- getStoredReferralDataFromSession,
- clearStoredReferralDataInSession,
- hasReferralDataExpired,

(Note: the "clearExpiredReferralData" is just a name that popped first in my mind. Perhaps there are better names for it.)

src/containers/AuthenticationPage/AuthenticationPage.helpers.js
```diff
@@ -35,22 +39,35 @@ export const getNonUserFieldParams = (values, userFieldConfigs) => {
...
+export const getExtendedDataMaybe = (
+  submitValues,
+  userTypeID,
+  userTypeConfigs,
+  userFields,
+  referralData
+) => {
+  const userTypeConfig = userTypeConfigs.find(config => {
+    return config.userType === userTypeID;
+  });
```
Gnito: I started to wonder if this could be made more useful/extendable.

The **userTypeConfigs** is only needed for **pickReferralData**, but that function does not pick anything from **_submitValues_**.

Perhaps the function should be something like this instead:

```js
export const getExtendedDataMaybe = (submitValues, userTypeId, userFields, extraData) => {
  const { publicData, privateData, protectedData } = extraData;

  return !isEmpty(submitValues)
    ? {
        publicData: {
         ...publicData,
          userType,
          ...pickUserFieldsData(submitValues, 'public', userTypeId, userFields),
        },
        privateData: {
          ...privateData,
          ...pickUserFieldsData(submitValues, 'private', userTypeId, userFields),
        },
        protectedData: {
          ...protectedData,
          ...pickUserFieldsData(submitValues, 'protected', userTypeId, userFields),
          // If the form has any additional values, pass them forward as user's protected data
          ...getNonUserFieldParams(submitValues, userFields),
        },
      }
    : {};
};
```

If it works like that, then the pickReferralData could happen earlier:
```js
const extraPrivateData = pickReferralData(userTypeConfig, referralData);
getExtendedDataMaybe(rest, userType, userTypes, userFields, { privateData: extraPrivateData }),
```

What do you think? Perhaps, that pattern would make it easier to include similar extra data in the future. I.e. avoid extra parameters and code changes for the _getExtendedDataMaybe_ function.

Gnito: E.g. one could easily add extra note to private data that user was using SSO when signing up.

Inside getHandleSubmitConfirm:
```js
const privateData = { ...pickReferralData(userTypeConfig, referralData), usedSSO: true };
// Pass other values as extended data according to user field configuration
const extendedDataMaybe = getExtendedDataMaybe(rest, userType, userFields, { privateData });
```

src/util/includeScripts.js
Amanda-Kauppinen: Sounds like a good idea! Now have `clearReferralDataIfExpired`. The first two of those functions are used elsewhere, but now `hasReferralDataExpired` is no longer exported, and checked that `hasReferralDataBeenStored` also doesn't need an export

src/containers/AuthenticationPage/AuthenticationPage.helpers.js
Amanda-Kauppinen: Yeah, I agree. It does stand out a bit that `pickReferralData` doesn't have anything to do with `submitValues`. Thanks for the example because I was unsure about duplicating the code, but that makes a lot of sense as a use case!

I also like that it wouldn't be so specific this way - then users could add any kind of data modifiers that make sense and this function will handle it without having to introduce risks by changing the logic of the default picking functionality.

src/containers/AuthenticationPage/AuthenticationPage.helpers.js
```diff
@@ -115,8 +139,18 @@ export const getHandleSubmitConfirm = ({ authInfo, submitSingupWithIdp, userFiel
...
+  const extraPrivateData = pickReferralData(userTypeConfig);
+
   // Pass other values as extended data according to user field configuration
-  const extendedDataMaybe = getExtendedDataMaybe(rest, userType, userFields, true);
+  const extendedDataMaybe = getExtendedDataMaybe(
+    rest,
+    userType,
+    userFields,
+    { privateData: extraPrivateData },
+    true
```
Gnito: Btw. it sounds like this true is some old leftover parameter which is not used anymore?

Amanda-Kauppinen: @Gnito I was going to ask about this as I think it came from the refactor authentication page branch. I just checked and [it's in the main branch also](https://github.com/sharetribe/web-template/blob/7d5e4efe159827070334174fe0cd8456e4d20c9d/src/containers/AuthenticationPage/AuthenticationPage.helpers.js#L119)

src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,195 @@
...
+    // If any key or value doesn't match, the stored data differs from the URL referral
+    if (
+      urlReferralKey !== storedDataKey ||
+      urlReferralParams[urlReferralKey] != storedData.sources[storedDataKey]
+    ) {
+      return false;
+    }
+  }
+
+  return true;
```
Gnito: There are multiple ways to do this kind of simple object comparisons.

Perhaps, the most common parttern is to turn them into strings. Since we know it's just key-value pairs, we could stringify these (with JSON.stringify) and compare strings:

```js
// Sort properties according to their string representation 
// (otherwise the order would be following the insertion order).
const sortedURLReferralParams = Object.fromEntries(Object.entries(urlReferralParams).sort());
const sortedStoredDataParams = Object.fromEntries(Object.entries(storedData?.sources).sort());
return JSON.stringify(sortedURLReferralParams) === JSON.stringify(sortedStoredDataParams);
```

Or the same with URLSearchParams:
```js
// Sort properties according to their string representation 
// (otherwise the order would be following the insertion order).
const searchParamsFromURL = new URLSearchParams(urlReferralParams);
const searchParamsFromStore = new URLSearchParams(storedData?.sources);
searchParamsFromURL.sort();
searchParamsFromStore.sort();
return searchParamsFromURL.toString() === searchParamsFromStore.toString();
```

src/util/sessionStorageHelpers.js
```diff
@@ -0,0 +1,195 @@
...
+  // Don't store the referral ID if it has already been stored and hasn't expired
+  if (!isReferralDataStored(storedData, referralData) || isReferralDataExpired(storedData)) {
+    const isBrowser = typeof window !== 'undefined';
+
+    if (isBrowser && window?.sessionStorage) {
+      // Convert referral ID values to strings before storing, this is because
+      // parse() parses the ID value into a number
+      const sources = Object.fromEntries(
+        Object.entries(referralData).map(([k, v]) => [k, typeof v === 'number' ? v.toString() : v])
+      );
```
Gnito: This could be done inside the replacer (to avoid doing the same thing in 2 different ways).

```js
const obj = { foo: 'bar', asdf: 1 };
const replacer = function(k, v) {
  if (typeof v === 'number') {
    return v.toString();
  }
  return v;
};
JSON.stringify(obj, replacer)
// => '{"foo":"bar","asdf":"1"}'

Gnito: Note: reviver does not need to restore the type back to number (if we don't want that).

Gnito (approved):
LGTM!

Gnito (approved):
LGTM!

PR #830, opened 2026-04-16
Attach file to message

SariSaar:
This PR adds capabilities to upload and download digital files to Sharetribe. This feature is enabled for marketplaces by default, and operators can disable the functionality for the entire marketplace in Console Access control.

As the first use case for file uploads and downloads, we're adding the capability to attach up to 10 files to a transaction message. This functionality is enabled per listing type, and it is disabled for existing listing types by default.

Uploading a file happens in a few steps:
- retrieve the file metadata
- use the metadata to create an `ownFile` entity
- use the `ownFile` entity id to create a signed upload storage URL associated with the `ownFile` entity
- upload the file to the storage URL outside the Sharetribe API
After the upload completes, the `ownFile` entity holds the reference to the uploaded file. 

To attach a file to a message, you'll need to include the `ownFile` entity id in a new message attribute array `publicFileAttachments` when sending the message.

Once a file has been uploaded, it can be sent. After upload, the file is still verified by the Sharetribe backend. If the verification succeeds for all files in a message, the message and the files are shown in the recipient's UI. A sent message with any files pending verification will show up in the sender's UI as greyed out, and verification status is shown for each file separately.

If a file's verification fails before the user sends the message, or the file upload encounters some other error, the sender must remove the file from the message before the message can be sent. The sender can also remove any other uploaded files from the message, as well as download a successfully uploaded file, before sending.

An uploaded file that has been removed from the message still exists in storage, but if it is not attached to a message, it is eventually removed by Sharetribe's backend automatically.

After a message has been sent, neither the sender or the recipient can delete a file. Only operators can delete files in Console after they have been sent in a message.

This PR also includes some refactorings that are good to note:
  - Message and OwnMessage components have been moved from ActivityFeed.js to Message.js
  - SendMessageForm functionality was managed under TransactionPanel, and it is now passed as a prop from TransactionPage to TransactionPanel similarly to ActivityFeed. This means that TransactionPanel remains more of a display component with less functionality
  - TransactionPage was refactored away from using Redux mapStateToProps and mapDispatchToProps functions, which have been deprecated.
   
### New translations

```
  "FileUpload.finishingUpload": "Finishing upload…",
  "FileUpload.maxFileSizeExceeded": "File too large (max 1 GB).",
  "FileUpload.mimeTypeError": "Upload failed. This file type isn't supported.",
  "FileUpload.removeFile": "Remove file {fileName}",
  "FileUpload.uploadFailed": "Upload failed. Please try again.",
  "FileUpload.uploading": "Uploading… Keep this tab open.",
  "FileUpload.verificationFailed": "Security check failed. Please contact support.",
  "FileUpload.verifying": "Scanning the file…",
  "Message.downloadFile": "Download file {fileName}",
  "Message.fileDeleted": "Deleted file.",
  "Message.fileSecurityCheckFailed": "Security check failed. Please contact support.",
  "Message.fileVerifying": "Scanning the file…",
  "Message.pendingVerificationNote": "This message will be sent after the security check is complete.",
  "Message.securityCheckFailedNote": "This message was not sent. One or more files failed a security check.",
  "TransactionPage.attachFile": "Attach a file",
  "TransactionPage.messageFilesDisabled": "{marketplaceName} has temporarily disabled file uploads and downloads. Contact support for more information.",
  "TransactionPage.navigationBlockedBeforeFilesSent": "You have unsent files. If you leave now, your uploads will be lost.",
```


src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,180 @@
+import React from 'react';
+import { useIntl, FormattedMessage } from 'react-intl';
+import classNames from 'classnames';
+import { calculateFileSize } from '../../util/fileHelpers';
+import { FileName } from '../../components';
+
+import css from './FileUpload.module.css';
+
+const IconCheck = () => (
```
Gnito: Perhaps, all this Icon* components could be moved to their own files (inside the same directory).

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,180 @@
...
+
+  let statusIcon;
+  if (uploadInProgress || verificationInProgress) {
+    // TODO include progress functionality
+    statusIcon = <IconSpinner />;
+  } else if (hasError) {
+    statusIcon = <IconError />;
+  } else if (hasCompletedUpload) {
+    statusIcon = <IconCheck />;
+  }
```
Gnito, line 66: This could be turned into its own component inside this file: `<StatusIcon {/* relevant props here*/} />`

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,180 @@
...
+      ) : error.message ? (
+        error.message
+      ) : (
+        <FormattedMessage id="FileUpload.uploadFailed" />
+      );
+  } else if (hasCompletedUpload && !verificationInProgress) {
+    const { size: sizeRaw } = file?.attributes;
+    const { size, unit } = calculateFileSize(sizeRaw);
+    statusText = `${size} ${unit}`;
+  }
```
Gnito, line 112: This could be turned into its own component inside this file: `<StatusText {/* relevant props here*/} />`

src/containers/TransactionPage/Message/Message.module.css
```diff
@@ -0,0 +1,187 @@
...
+  gap: var(--spacing-lg, 12px);
+
+  @media (--viewportMedium) {
+    padding: 8px 14px 8px 14px;
+    margin: 0;
+  }
+}
+
+.messageContent {
+  flex: 1;
```
Gnito: Minor detail about personal preferences:

This "flex: 1" is a shorthand that might be a bit difficult to remember what it stands for. 
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex

I have the habit of always using `flex-grow: 1;` instead. As that more clearly gives a hint what's happening (it affects growing, not shrinking or base-width).
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-grow

There are other shorthands that are more clear (e.g. `margin: 0px 12px 0 0;`), but there those different shorthand values are more of a siblings to each other, while `flex: 1;` and `flex: 1em;` are affecting completely different kind of values.

So, perhaps, you could change all the `flex: 1;` rules on this PR to `flex-grow: 1;`. It seems that there are 4 of them.

src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -33,6 +34,8 @@ const { UUID } = sdkTypes;
 const MESSAGES_PAGE_SIZE = 100;
 const REVIEW_TX_INCLUDES = ['reviews', 'reviews.author', 'reviews.subject'];
 const MINUTE_IN_MS = 1000 * 60;
+const POLL_MAX_ATTEMPTS = 60; // TODO check with the team for a reasonable max attempt count
+const POLL_INTERVAL_MS = 1000;
```
Gnito, line 38: One pattern with these kind of polling setup is using variable length interval. 
- if there's some reason why server is under load, we could start making calls more rarely
- if user has already waited for 15 seconds, it doesn't really make that much difference to wait similar amount of time for the next attempt. 

To illustrate: first 3 attempts with 5 seconds, next 3 with 15 seconds, next 3 with 30 seconds.

Or perhaps delay that increases on each attempt?

```js
// copy-paste to browser's console
const delay = ms => new Promise(resolve => window.setTimeout(resolve, ms));
const mockXHRCall = () => delay(1000).then(() => ({ data: 'data' }));
const attempts = 5;
const withRetry = (xhrCall, attempt, delayMs) => {
  console.log('Calling API...');
  xhrCall()
    .then(() => {
      // failed
      console.log('  Attempt', attempt, 'of', attempts, 'failed.');
      if (attempt >= attempts) {
        console.log('Max attempts reached. Giving up.');
        return;
      }
      console.log('  Retrying in', delayMs, 'ms.');
      return delay(delayMs).then(() => withRetry(xhrCall, (attempt + 1), (delayMs * 2)));
    });
};
withRetry(mockXHRCall, 1, 2000);
```

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -75,6 +75,10 @@ import {
   fetchMoreMessages,
   fetchTimeSlots,
   fetchTransactionLineItems,
+  uploadFile,
+  clearUploadedFiles,
+  selectFileUploads,
+  downloadFile,
 } from './TransactionPage.duck';
 import css from './TransactionPage.module.css';
 import { getCurrentUserTypeRoles, hasPermissionToViewData } from '../../util/userHelpers.js';
```
Gnito: Could you move this to correct place aka next to other util files.

src/components/FileName/FileName.module.css
```diff
@@ -0,0 +1,19 @@
+.root {
+  display: flex;
+  flex: 1;
+  min-width: 0;
+  overflow: hidden;
+}
+
+.baseName {
+  flex-shrink: 1;
+  min-width: 0;
```
Gnito: This sounds strange. Is it needed?

src/components/FileName/FileName.module.css
```diff
@@ -0,0 +1,19 @@
...
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+
+.extension {
+  flex-shrink: 0;
+  white-space: nowrap;
+}
```
Gnito: If we end up moving ellipsis towards the middle of the name, then this naming pattern should be changed.
I wonder if there could be something else already: e.g. `.dynamicNameBeginning` & `.fixedNameEnding`

src/components/FileUpload/FileUpload.module.css
```diff
@@ -0,0 +1,95 @@
+.root {
+  display: flex;
+  padding: 12px;
+  align-items: center;
+  gap: 8px;
+  align-self: stretch;
```
Gnito: No need to change, but the order of these could be considered a bit more.

```css
/* This is an exceptional rule for the layout imposed by the parent element,
which uses display: flex; or display: grid; */
align-self: stretch;

/* These are all flexbox-related layout rules for this element */
display: flex;
align-items: center;
gap: 8px;

/* Padding tells about white-space inside this current container */
padding: 12px;
```


- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-self#examples
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/gap

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const rootClass = classNames(className || css.root, hasError && css.error);
+  const fileInfoClass = classNames(css.fileInfo, hasError && css.fileInfoError);
```
Gnito: - Let's stick to our className vs rootClassName pattern
- The classNames library has its own syntax to add optional classes. Let's use that (the consistency makes it easier to follow the existence of those different conditional classes).


```suggestion
  const classes = classNames(rootClassName || css.root, className, { [css.error]: hasError });
  const fileInfoClass = classNames(css.fileInfo, { [css.fileInfoError]: hasError });
```

src/containers/TransactionPage/Message/Message.js
```diff
@@ -0,0 +1,323 @@
...
+  });
+};
+
+const FileAttachment = props => {
+  const { fileAttachment, downloadFile, intl } = props;
+  const { file } = fileAttachment;
+  const isAvailable = file.attributes.state === 'available' && fileAttachment.file;
+  const isVerifying = file.attributes.state === 'pendingVerification';
+  const isVerificationFailed = file.attributes.state === 'verificationFailed';
+  // TODO double check deletion conditions from file attribute when deletion implemented in Core!
```
Gnito: Just highlighting this TODO

src/containers/TransactionPage/Message/Message.js
```diff
@@ -0,0 +1,323 @@
...
+  const isDeleted = !file || file.attributes.deleted;
+
+  const name = file?.attributes?.name;
+
+  const rootClass = classNames(css.fileAttachment, isAvailable && css.fileAttachmentAvailable);
+
+  if (isAvailable) {
+    const { size, unit } = calculateFileSize(file.attributes.size);
+    return (
+      <div className={rootClass} onClick={() => downloadFile(fileAttachment.id)}>
```
Gnito: If this works as a button, you should make it a button (improve accessibility by using semantic HTML).

src/containers/TransactionPage/Message/Message.js
```diff
@@ -0,0 +1,323 @@
...
+        <span
+          role="img"
+          aria-label={intl.formatMessage({ id: 'Message.downloadFile' })}
```
Gnito: These are on wrong element. The role attribute should be given for `<svg>` element.

And I think it should be `role="none"` without aria-label. 

If this is turned into a semantic element (button) then the aria-label is relevant for that active element (if needed) - and if it's not needed, then the child elements would be given for the screenreader => which means that this icon would be read + filename + file size). There should not be repetition - which is somewhat unavoidable if there are multiple different elements contributing to the content of the button.

src/containers/TransactionPage/Message/Message.js
```diff
@@ -0,0 +1,323 @@
...
+      </div>
+    );
+  }
+
+  if (isVerifying) {
+    return (
+      <div className={rootClass}>
+        <span
+          role="img"
+          aria-label={intl.formatMessage({ id: 'Message.fileVerifying' })}
```
Gnito: I think it should be `role="none"` without aria-label and the role needs to be given to `<svg>` element.

(and let's not repeat things for screenreaders: 
`<div>Scanning the file... [file name] Scanning the file...</div>`)

src/util/fileHelpers.js
```diff
@@ -0,0 +1,24 @@
+const KILO = 1024;
+
+export const MAX_FILE_SIZE = KILO * KILO * KILO;
+
+export const calculateFileSize = size => {
+  const kbSize = Math.ceil(size / KILO);
+  const mbSize = (kbSize / KILO).toFixed(1);
```
Gnito: Leave this "toFixed(1)" for the React Intl.

src/util/fileHelpers.js
```diff
@@ -0,0 +1,24 @@
...
+
+  const useMb = mbSize > 1;
+
+  return useMb ? { size: mbSize, unit: 'MB' } : { size: kbSize, unit: 'KB' };
```
Gnito: Perhaps it's better that the unit is in small letters (kb, mb) to be passed as arguments for the React Intl's "select". 🤔

src/util/fileHelpers.js
Gnito: Or rather, we could let React Intl to define the precision of the number.

I.e. the important part here is to define when different units are used.
Do we have 110 KB or 0.1 MB? (or 110 KB vs 0.11 MB - if 2 decimals are allowed)

src/util/fileHelpers.js
Gnito: I think we should add some tests to this function as it's something that we/customizers might want to tune and it would be good to have a collection of test numbers.

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,180 @@
...
+    <div className={rootClass}>
+      <span className={css.statusIcon}>{statusIcon}</span>
+      <div className={fileInfoClass}>
+        {fileNameElement}
+        <span className={css.statusText}>{statusText}</span>
+      </div>
+      <button
+        className={css.removeButton}
+        onClick={() => onRemoveFile(tempId)}
+        aria-label={intl.formatMessage({ id: 'FileUpload.removeFile' })}
```
Gnito: From accesibility point of view: this might need the filename as an argument.
(at least some tools warn if page contains multiple active components that have the exact same text, but they perform actions to different files/listings/users)

src/containers/TransactionPage/Message/Message.module.css
SariSaar: Good call, updated now!

src/containers/TransactionPage/Message/Message.js
SariSaar: Changed into a button and moved the aria-label to button instead of the span!

src/components/FileUpload/FileUpload.js
SariSaar: Good point! Added file name as argument for both file removal here and file download in Message

src/containers/TransactionPage/TransactionPage.duck.js
SariSaar: Added the increase on each attempt logic to both polling functions now. Still need to determine the attempt count so I left it as TODO for now.

src/util/fileHelpers.js
SariSaar: Updated to use Intl.NumberFormat

src/util/fileHelpers.js
SariSaar: I added some tests here!

src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -467,15 +475,274 @@ export const fetchMoreMessages = (txId, config) => (dispatch, getState, sdk) =>
...
+      if (fileState === 'verificationFailed') {
+        const [fileUpload] = denormalisedResponseEntities(resp);
+        return rejectWithValue({ tempId, message: 'verificationFailed', fileUpload });
+      }
+
+      if (attempt >= POLL_MAX_ATTEMPTS - 1) {
+        return rejectWithValue({ tempId, message: 'timeout' });
+      }
+
+      return delay(delayMs).then(() => poll(attempt + 1, delayMs * 2));
```
Gnito: My previous suggestion was just that: suggestion. If we think this makes sense, then we could do something like "increase polling interval up until point X", which could be 60 seconds, after which we just poll every 60 seconds.


```suggestion
      const currentDelayMs = delayMs < 600000 ? delayMs * 2 : 60000;
      return delay(delayMs).then(() => poll(attempt + 1, currentDelayMs));
```

Another thing that could be considered is checking `document.hidden`, but I'm not sure if it makes sense.

src/components/FileName/FileName.module.css
SariSaar: Apparently not 👍 cleared!

src/components/FileName/FileName.module.css
SariSaar: Good point, changed this now!

server/csp.js
```diff
@@ -13,6 +13,7 @@ const baseUrl = process.env.REACT_APP_SHARETRIBE_SDK_BASE_URL || 'https://flex-a
 // cdn.st-api.com
 // If assetCdnBaseUrl is used to initialize SDK (for proxy purposes), then that URL needs to be in CSP
 const assetCdnBaseUrl = process.env.REACT_APP_SHARETRIBE_SDK_ASSET_CDN_BASE_URL;
+const fileStorageDomain = process.env.REACT_APP_FILE_STORAGE_DOMAIN;
```
Gnito: - Do we need REACT_APP_FILE_STORAGE_DOMAIN inside the React app?
  - Does this need to be prefixed with "REACT_APP_" (now or in the future?)

- If we don't have default domain baked in somewhere, this means that the REACT_APP_FILE_STORAGE_DOMAIN becomes mandatory environment variable and it needs to be asked through scripts/config.js
  - E.g. REACT_APP_SHARETRIBE_SDK_ASSET_CDN_BASE_URL has the default baked into the SDK
  - baseUrl has the default added more explicitly

SariSaar: - Changed the env variable to server-side only – I couldn't think of a use case where we would use this client-side, good point!
- Added the variable prompt to scripts/config.js

src/components/FileName/FileName.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { analyseFileName } from '../../util/fileHelpers';
+
+import css from './FileName.module.css';
+
+const FileName = ({ name, className }) => {
+  const { baseName, extension } = analyseFileName(name);
+  return (
+    <span className={className || css.root}>
```
Gnito: Since this is shared component, perhaps it would make sense to follow our class pattern here too:

`const classes = classNames(rootClassName || css.root, className);`

src/components/FileUpload/FileUpload.js
```diff
@@ -0,0 +1,135 @@
...
+      );
+  } else if (uploadState === FILE_UPLOAD_COMPLETE) {
+    const { size } = file?.attributes;
+    statusText = calculateFileSize(size, locale);
+  }
+
+  return <span className={css.statusText}>{statusText}</span>;
+};
+
+const FileUpload = props => {
```
Gnito, line 139: Add JSDoc comments (since this is a shared component, this helps people who use code editors)

src/components/FileName/FileName.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { analyseFileName } from '../../util/fileHelpers';
+
+import css from './FileName.module.css';
+
+const FileName = ({ name, className }) => {
```
Gnito: Add JSDoc comments (since this is a shared component, this helps people who use code editors)

src/containers/TransactionPage/SendMessageForm/SendMessageForm.js
```diff
@@ -90,11 +119,37 @@ class SendMessageFormComponent extends Component {
             invalid,
             form,
```
Gnito: This is old code, but could you rename this `form: formApi,`
It's pretty confusing what the "form" does and and it's also difficult to find with text search. So, I've been trying to convert these to formApi.

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -214,6 +218,34 @@ const getDataValidationResult = (transaction, process) => {
...
+ */
+const useUploadNavigationBlock = (when, history, message) => {
+  useEffect(() => {
+    if (!when) return;
+
+    const unblock = history.block(message);
+
+    const handleBeforeUnload = e => {
+      e.preventDefault();
+      e.returnValue = '';
```
Gnito, line 244: I'm not sure if this is needed, but if used: add code comment:
// Included for legacy support, e.g. Chrome/Edge < 119

https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#examples

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -214,6 +218,34 @@ const getDataValidationResult = (transaction, process) => {
     : { hasValidData: true, errorMessageId: null };
 };
 
+/**
+ * Blocks React Router in-app navigation and browser-level navigation while `when` is true.
+ * Registers history.block() for React Router transitions and a beforeunload listener for
+ * browser-level events (refresh, tab close). Both are cleaned up when `when` becomes false.
+ *
+ * @param {boolean} when - Whether to block navigation
```
Gnito: I'm not sure if this is the clearest name. 🤔

src/containers/TransactionPage/TransactionPanel/TransactionPanel.js
```diff
@@ -131,16 +137,40 @@ export class TransactionPanelComponent extends Component {
...
+    if (file) {
+      const tempId = `${Date.now()}-${Math.random()}`;
+
+      this.props.onUploadFile(file, tempId);
+    }
+  }
+
+  onRemoveFile(tempId) {
+    this.props.onClearUploadedFiles([tempId]);
+  }
```
Gnito: I'm not sure if it makes sense to have these functions here.
If you have time, you could perhaps consider/refactor this a bit more.

TransactionPanel is not really something that needs to know about files. 

One option for now/future is to avoid drop-drilling by moving the creation of `<SendMessageForm>` already on TransactionPage component. (similarly to `<ActivityFeed>` et al.) In practice, that would also mean that other mehtods for this form would need to be extracted away from this component.

src/translations/en.json
```diff
@@ -655,6 +662,12 @@
   "ManageListingsPage.queryError": "Query failed. Please try again.",
   "ManageListingsPage.title": "Manage your listings",
   "ManageListingsPage.youHaveListings": "You have {count} {count, plural, one {listing} other {listings}}",
+  "Message.downloadFile": "Download file {fileName}",
+  "Message.fileDeleted": "Deleted file.",
+  "Message.fileSecurityCheckFailed": "Security check failed. Please contact support.",
+  "Message.fileVerifying": "Scanning the file...",
```
Gnito:
```suggestion
  "Message.fileVerifying": "Scanning the file…",
```

src/util/fileHelpers.js
```diff
@@ -0,0 +1,34 @@
+const KILO = 1024;
+
+export const MAX_FILE_SIZE = KILO * KILO * KILO;
```
Gnito: There's some confusion between unit systems (SI vs IEC):
https://en.wikipedia.org/wiki/Byte#Multiple-byte_units

You could perhaps verify that core uses 1024 (almost certainly they do).

Still, it wouldn't hurt to be more explicit here:
```suggestion
const KILO = 1024; // 2^10 aka kibibyte (KiB)
const GIGA = KILO * KILO * KILO; // 2^30 aka gibibyte (GiB)

export const MAX_FILE_SIZE = GIGA;
```

src/util/fileHelpers.js
```diff
@@ -0,0 +1,34 @@
...
+
+export const calculateFileSize = (size, locale) => {
+  const kbSize = Math.ceil(size / KILO);
+  const mbSize = kbSize / KILO;
```
Gnito, line 8: I wonder if we should add a comment like this (or is it causing more confusion):
`// Convert to kibibyte (2^10) and mebibyte (2^20) as Intl.NumberFormat does not support them natively.`

src/util/fileHelpers.js
SariSaar: Pinged Core in a comment in the wireframes

src/util/fileHelpers.js
Gnito: Hmm. Actually, it seems to be somewhat agnostic (there's just "compact" style that creates something funny 1000000000 => "1B GB".

Closing this issue.

src/util/fileHelpers.js
SariSaar: Walter confirmed that Core uses 1024

src/containers/TransactionPage/TransactionPanel/TransactionPanel.js
SariSaar: I made an attempt on the refactor now, any comments welcome!

src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -467,15 +486,290 @@ export const fetchMoreMessages = (txId, config) => (dispatch, getState, sdk) =>
...
+
+      if (attempt >= POLL_MAX_ATTEMPTS - 1) {
+        return rejectWithValue({ tempId, message: 'timeout' });
+      }
+
+      return delay(delayMs).then(() => poll(attempt + 1, getNextPollingDelay(delayMs)));
+    });
+  };
+
+  return poll(0, POLL_INTERVAL_MS).catch(e => rejectWithValue({ tempId, error: storableError(e) }));
```
lyyder: Is it now so that when there's an upload, `pollFirFileVerificationPayloadCreator` is instantly called? And it first fetches the current status, then applies the delay and then polls again?

If this is the case, then the first poll is likely useless, as it takes some time for the file verification to pass. It's a good idea that the delay is increased over time. However, we could consider adding an initial wait, that is longer than the first polling, say 2 seconds. After that we'd start the polling, but with a lower initial delay. So something like this:

1. Upload file
2. Apply initial wait, for 1000-2000 ms
3. Start polling, with 500 ms delay, multiplying by ~1.5-2 on each round

This way we would not lose the first one or two requests, where it is likely that the file is not verified yet. We could observe a bit, what is an average verification time for a file and adjust accordingly. Any thoughts on this?

lyyder (commented):
Looks good! A few questions, not super critical bot not clear to me:

1. If an upload fails, is there a possibility to reuse the same file entity to try the upload again?
2. If message sending fails, is there a possibility to try sending the message again without doing a re-upload?

src/containers/TransactionPage/TransactionPage.duck.js
SariSaar: I think that adding a longer delay in the beginning of upload file verification makes sense! For messages, we can start polling right away, because there's already a delay from sending the message. @Gnito what do you think?

Gnito: Sounds good!

SariSaar:
> Looks good! A few questions, not super critical bot not clear to me:
> 
>     1. If an upload fails, is there a possibility to reuse the same file entity to try the upload again?
> 
>     2. If message sending fails, is there a possibility to try sending the message again without doing a re-upload?

@lyyder great questions!
1. Currently not. There's nothing in the product spec that states this should be possible, so it's more of a product design question.
2. Currently not – also nothing specific in the spec about this. *edit I re-checked the spec and I think this actually should be possible, so I'll make the necessary changes!

I'll ask Juho what the functionality should be related to these two things!

SariSaar:
> > Looks good! A few questions, not super critical bot not clear to me:
> > ```
> > 1. If an upload fails, is there a possibility to reuse the same file entity to try the upload again?
> > 
> > 2. If message sending fails, is there a possibility to try sending the message again without doing a re-upload?
> > ```
> 
> @lyyder great questions!
> 
>     1. Currently not. There's nothing in the product spec that states this should be possible, so it's more of a product design question.
> 
>     2. Currently not – also nothing specific in the spec about this. *edit I re-checked the spec and I think this actually should be possible, so I'll make the necessary changes!
> 
> 
> I'll ask Juho what the functionality should be related to these two things!

Update: 
1. Juho said that this is not necessary
2. I now added logic that checks that message id exists before clearing the file uploads, so that a failed message can be retried with the same file uploads.

src/containers/TransactionPage/TransactionPage.duck.js
```diff
@@ -467,15 +486,290 @@ export const fetchMoreMessages = (txId, config) => (dispatch, getState, sdk) =>
...
+      throw new Error('Missing file, cannot initiate upload.');
+    }
+
+    if (Object.keys(getState().TransactionPage.fileUploads).length > MAX_FILE_UPLOAD_COUNT) {
+      throw new Error('Upload file count exceeded, cannot initiate upload.');
+    }
+
+    if (file.size > MAX_FILE_SIZE) {
+      throw new Error('File too large (max 1 GB).', file);
+    }
```
lyyder, line 570: @SariSaar If the max file size changes in the API, it will require an update here. Template could also rely on the error response from `sdk.ownFiles.create` to communicate too large file size. It will require one API call to fail but it's a lightweight call, it would act as a reference on how to react to file creation validation errors, and it would bundle the file size and mime type errors handling in the same place in the code. Not a major thing but something to consider still in this initial release or future releases.

lyyder (commented):
One more comment on file size validation.

src/containers/TransactionPage/TransactionPage.duck.js
SariSaar: Had a chat about this IRL, and we decided that we'll add a comment on this for customisers but won't change the logic for now.

Gnito (approved):
LGTM

PR #835, opened 2026-04-24
configHelpers: Reworked mergeSortConfig to better account for missing hosted sort options

Amanda-Kauppinen:
Fixes a bug where listing metadata fields were not shown in the sort dropdown if the default sorting config is missing


Gnito (approved):
LGTM!

PR #849, opened 2026-05-18
Initiate digital purchase

shareoc:


src/translations/en.json
```diff
@@ -178,11 +180,24 @@
...
   "DiscardDraftModal.message": "This can't be undone.",
   "DiscardDraftModal.submit": "Discard draft",
   "DiscardDraftModal.title": "Are you sure you want to discard this draft?",
   "DisputeModal.close": "Close",
+  "ReportModal.close": "Close",
+  "ReportModal.description": "Tell us what went wrong. We'll review the issue and get back to you as soon as possible.",
+  "ReportModal.reportPlaceholder": "There's a problem with...",
+  "ReportModal.reportReasonRequired": "Please tell us what went wrong",
+  "ReportModal.reportSubmitFailed": "Oops, something went wrong. Please refresh the page and try again.",
+  "ReportModal.label": "What went wrong?",
```
Gnito: Remember to sort this file alphabetically

Gnito (commented):
**_/api/delete-account_** needs to be handled!

I think you should add also code-comment to transaction.js
https://github.com/sharetribe/web-template/blob/main/src/transactions/transaction.js#L19
"If you add new processes, remember to check rules on account deletion on the server" or something like that.

And something similar could be mentioned on README.md on that directory:
https://github.com/sharetribe/web-template/blob/main/src/transactions/README.md

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -792,6 +829,7 @@ export const TransactionPageComponent = props => {
     isNegotiationProcess,
     isBookingProcess: isBookingProcess(processName),
     isPurchaseProcess: processName === PURCHASE_PROCESS_NAME,
+    isDownloadProcess: isDownloadProcess(processName),
     isInquiryProcess: processName === INQUIRY_PROCESS_NAME,
     isRegularNegotiation,
```
Gnito: Let's make this setup more extendable. Pass in the processName and detect the process inside _TransactionFields.js_ instead.

```suggestion
    processName,
    isRegularNegotiation,
```

src/components/OrderPanel/OrderPanel.js
```diff
@@ -319,7 +326,8 @@ const OrderPanel = props => {
   const isBooking = isBookingProcess(processName);
   const isPurchase = isPurchaseProcess(processName);
   const isNegotiation = isNegotiationProcess(processName);
-  const isPaymentProcess = isBooking || isPurchase || isNegotiation;
+  const isDownload = isDownloadProcess(processName);
+  const isPaymentProcess = isBooking || isPurchase || isNegotiation || isDownload;
```
Gnito, line 330: Note to self: we probably want to push this "isPaymentProcess" information to transaction process (transaction.js) at some point.

src/config/configListing.js
```diff
@@ -282,7 +282,38 @@ export const listingFields = [
  *                          configuration may still disable uploading and downloading files, even if enabled in the listing type.
  */
 
+// TODO: Remove before merge
```
Gnito: This comment is just a reminder.

(Btw. probably, we just want to comment this out and move the lines to the bottom of the "listingTypes" array.)

src/containers/CheckoutPage/StripePaymentForm/StripePaymentForm.js
```diff
@@ -570,6 +571,8 @@ class StripePaymentForm extends Component {
       const checked = event.target.checked;
       this.updateBillingDetailsToMatchShippingAddress(checked);
     };
+    const isBooking = processName === BOOKING_PROCESS_NAME;
+    const isDownload = processName === DOWNLOAD_PROCESS_NAME;
```
Gnito: Use **_isBookingProcess_** and **_isDownloadProcess_** functions. That way someone could more easily have processes that are almost the same mapped inside these generic functions instead of spreading the exact process names to multiple components.

src/containers/CheckoutPage/CheckoutPage.js
```diff
@@ -137,21 +122,6 @@ const EnhancedCheckoutPage = props => {
       listing,
     });
     return <NamedRedirect name="ListingPage" params={params} />;
```
Gnito: Perhaps, this redirection should also happen in CheckoutPageAccessWrapper.

It kind of feels more straightforward that redirections back to listing page (aka access was denied) are handled on the access wrapper.

src/containers/InboxPage/InboxPage.stateDataDownload.js
```diff
@@ -0,0 +1,49 @@
...
+
+  return new ConditionalResolver([processState, transactionRole])
+    .cond([states.INQUIRY, _], () => {
+      return { processName, processState, actionNeeded: true };
+    })
+    .cond([states.PENDING_PAYMENT, CUSTOMER], () => {
+      return { processName, processState, actionNeeded: true };
+    })
+    .cond([states.PENDING_PAYMENT, PROVIDER], () => {
+      return { processName, processState };
```
Gnito: Btw. there's `.default()` condition that handles the default cases that just return processName and processState.

```js
    .default(() => {
      return { processName, processState };
    })
```

src/containers/TransactionPage/ReportModal/ReportModal.js
```diff
@@ -0,0 +1,165 @@
...
+      const classes = classNames(rootClassName || css.formRoot, className);
+      const submitInProgress = reportInProgress;
+      const submitDisabled = invalid || disabled || submitInProgress || reportSubmitted;
+
+      return (
+        <Form className={classes} onSubmit={handleSubmit}>
+          <FieldTextInput
+            className={css.reportMessage}
+            type="textarea"
+            id={formId ? `${formId}.reportReason` : 'reportReason'} //
```
Gnito: Remove the "//"

src/containers/TransactionPage/TransactionPanel/DiminishedActionButtonMaybe.js
```diff
@@ -11,17 +11,30 @@ import css from './TransactionPanel.module.css';
...
-  const { id, className, rootClassName, showDispute, onOpenDisputeModal } = props;
+  const {
+    id,
+    className,
+    rootClassName,
+    showDispute,
+    onOpenDisputeModal,
+    showReport,
+    onOpenReportModal,
+  } = props;
```
Gnito:
```suggestion
  const {
    id,
    className,
    rootClassName,
    showButton,
    buttonMessage,
    onOpenModal,
  } = props;
```

Then there could be 

```js
  const diminishedActionButton =
    showButton && buttonMessage && onOpenModal ? (
      <InlineTextButton id={id} className={css.diminishedActionButton} onClick={onOpenModal}>
        {buttonMessage}
      </InlineTextButton>
    ) : null;
```

src/containers/TransactionPage/TransactionPanel/PanelHeading.js
```diff
@@ -83,7 +84,7 @@ const PanelHeading = props => {
         <p className={css.transactionInfoMessage}>
           <FormattedMessage
             id={`TransactionPage.${processName}.${transactionRole}.${processState}.extraInfo`}
-            values={{ customerName, providerName, deliveryMethod, breakline }}
+            values={{ customerName, providerName, marketplaceName, deliveryMethod, breakline }}
```
Gnito, line 87: This feels wrong. If we don't have this on purchase process, why we have it here?

I'll ask from Juho&Katri through wireframes.

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -498,8 +522,16 @@ export const TransactionPageComponent = props => {
     const { reviewRating, reviewContent } = values;
     const rating = Number.parseInt(reviewRating, 10);
     const { states, transitions } = process;
+
+    // The download process only supports a single customer-side review transition.
+    // Bidirectional review transitions (REVIEW_1/REVIEW_2) don't exist in that process.
     const transitionOptions =
-      transactionRole === CUSTOMER
+      processName === DOWNLOAD_PROCESS_NAME
```
Gnito:
```suggestion
      isDownloadProcess(processName)
```

src/containers/TransactionPage/TransactionPage.js
```diff
@@ -64,7 +66,9 @@ import RequestQuote from './RequestQuote/RequestQuote';
 import Offer from './Offer/Offer';
 import TransactionFields from './TransactionFields/TransactionFields.js';
 import ActivityFeed from './ActivityFeed/ActivityFeed';
+import FileDownloadList from './FileDownloadList/FileDownloadList';
```
Gnito: Our API uses term "file attachements". So, I think this should follow that naming pattern.

In addition, Sari has already made a <FileAttachment> component inside TransactionPage/Message/Message.js. In that file, there's also handling for the access-control feature. 

I think we should now try to extract the relevant code into a new component: `TransactionPage/FileAttachements/FileAttachements.js`

- https://github.com/sharetribe/web-template/blob/initiate-digital-purchase/src/containers/TransactionPage/Message/Message.js#L171-L182

- => `<FileAttachements>`
  Naturally it would need to get the list of correct file attachments through props. 
  - Message: `fileAttachments={publicFileAttachments}`
  - Transaction: `fileAttachments={protectedFileAttachments}`

  Note: some kind of "**allowFiles**" flag needs to come from props too because requirements for file attachements for messages and transactions are different. For transactions, the requirement only comes from accesscontrol asset (but messages need to check listing type too).
  - `const allowFilesFromAccessControl = !config.accessControl.marketplace.fileUploadAndDownloadDisabled;`

src/translations/en.json
```diff
@@ -1196,6 +1220,11 @@
...
   "TransactionPage.ActivityFeed.default-booking.reviewed-by-customer": "{actor, select, you {You left a review for {otherUsersName}.} other {{otherUsersName} left a review for you. {reviewLink}}}",
   "TransactionPage.ActivityFeed.default-booking.reviewed-by-provider": "{actor, select, you {You left a review for {otherUsersName}.} other {{otherUsersName} left a review for you. {reviewLink}}}",
+  "TransactionPage.ActivityFeed.default-download.canceled": "The purchase was canceled.",
+  "TransactionPage.ActivityFeed.default-download.completed": "{transactionRole, select, customer {Your purchase was completed. {reviewLink}} other {{otherUsersName}'s purchase was completed.}}",
+  "TransactionPage.ActivityFeed.default-download.purchased": "{actor, select, you {You purchased {listingTitle}.} other {{otherUsersName} purchased {listingTitle}.}}",
+  "TransactionPage.ActivityFeed.default-download.reported": "{actor, select, you {You reported a problem.} other {{otherUsersName} reported a problem.}}",
+  "TransactionPage.ActivityFeed.default-download.reviewed": "{actor, select, you {You left a review.} other {{actor} left a review.}}",
```
Gnito: @shareoc this is wrong. There needs to be translation for

Wireframes had this note:
> Implementation notes:
If review period expired, activity feed copy is “Review period ended.”

https://github.com/sharetribe/web-template/blob/main/src/containers/TransactionPage/ActivityFeed/ActivityFeed.js#L142

I have not tested this, but perhaps something like would do the job:
```suggestion
  "TransactionPage.ActivityFeed.default-download.reviewed": "{actor, select, system {Review period ended.} you {You left a review.} other {{otherUsersName} left a review.}}",
```

src/containers/TransactionPage/TransactionPage.js
Gnito: AllowFiles would also need to take this `isProblemReported && isCustomerRole` into account.

src/containers/TransactionPage/TransactionPage.js
shareoc: @Gnito Which approach should we use to identify the process?


`isBookingProcess: processName === BOOKING_PROCESS_NAME`
or
`isBookingProcess: isBookingProcess(processName)`

EDIT: nevermind, you addressed this in the next comment..

src/containers/CheckoutPage/CheckoutPage.js
shareoc: @Gnito Valid, but the redirects in CheckoutPageAccessWrapper deal with authorization, whereas this deals with invalid data. Also wouldn't that require moving isDataLoaded into the wrapper which feels a bit awkward?

src/containers/TransactionPage/TransactionPanel/PanelHeading.js
shareoc: Moved the discussion to Slack here: https://sharetribe.slack.com/archives/C0B6RQ3536X/p1780326418374519

src/containers/TransactionPage/FileAttachments/FileAttachments.js
```diff
@@ -0,0 +1,194 @@
...
+  if (allowFiles && !fileAttachments?.length) {
+    return null;
+  }
+
+  const classes = classNames(rootClassName || css.container, className);
+
+  return (
+    <div className={classes}>
+      <Heading as="h3" rootClassName={css.sectionHeading}>
+        File(s)
```
thomasmalbaux: This probably should not be hardcoded.

src/translations/en.json
```diff
@@ -178,11 +180,23 @@
...
   "DeleteAccountForm.resetPasswordLinkText": "Send reset instructions.",
   "DeleteAccountForm.stripeDeletionFailedError": "Account deletion failed. Stripe data could not be deleted. Please contact support.",
+  "DigitalDownloadForm.breakdownTitle": "Order breakdown",
+  "DigitalDownloadForm.ctaButton": "Buy now",
   "DiscardDraftModal.close": "Close",
   "DiscardDraftModal.message": "This can't be undone.",
   "DiscardDraftModal.submit": "Discard draft",
   "DiscardDraftModal.title": "Are you sure you want to discard this draft?",
   "DisputeModal.close": "Close",
+  "ReportModal.close": "Close",
```
Gnito: @shareoc re-order this file alphabetically.

src/components/OrderPanel/OrderPanel.js
```diff
@@ -597,6 +616,8 @@ const OrderPanel = props => {
               <FormattedMessage id="OrderPanel.ctaButtonMessageMakeOffer" />
             ) : showRequestQuoteForm ? (
               <FormattedMessage id="OrderPanel.ctaButtonMessageRequestAQuote" />
+            ) : showDownloadForm ? (
+              <FormattedMessage id="DigitalDownloadForm.ctaButton" />
```
Gnito: @shareoc We should not use this one. This button is used to open the order form on mobile layout and it has different requirements than form's own submit button. I.e. all these "OrderPanel.ctaButton*" are meant to be distinct translations so that operators can use some "short" text on button that's fixed to the bottom of the mobile layout. Meanwhile the form's own submit can be longer (it does not have fixed positioning and therefore it can be multiline button).

src/translations/en.json
Gnito: this was done on the base branch

PR #852, opened 2026-05-19
configHelpers.js: Only validate numberConfig for schemaType long

SariSaar:
Fix an issue where a Console-created config has an obsolete numberConfig defined for a non-numeric schema type.


Gnito (approved):
LGTM!

PR #853, opened 2026-05-19
Add files to listings

Amanda-Kauppinen:
Adds a files tab to the `EditListing` wizard for listing types with file attachments enabled. Handles the full upload lifecycle - progress tracking, security verification polling, and attaching files to the listing via `protectedFileAttachments`. Enforces file count and size limits, blocks publishing until all files are verified, takes into account global access control settings, and preserves upload state across tab navigations with more specific handling for draft vs. published listings.


src/util/configHelpers.js
```diff
@@ -1116,6 +1116,10 @@ export const allowProviderUpdateOffer = listingTypeConfig => {
   return listingTypeConfig?.negotiationOptions?.providerUpdateOffer === true;
 };
 
+export const requireListingFiles = listingTypeConfig => {
+  return listingTypeConfig?.messagingOptions?.fileAttachments === true;
```
Gnito: This is **messagingOptions** => it affects messages only.

Check the `requireListingImage` for reference how `defaultListingFields` works. 
(and move this helper function above it.)

Amanda-Kauppinen: Ah, I didn't read closely enough in Console that the toggle was for messages for listings. I'll need to mock it until it's ready then (I thought it was already, but the option is not visible in console at least)

Amanda-Kauppinen: ✅

src/util/configHelpers.js
```diff
@@ -1093,6 +1093,11 @@ export const displayDeliveryShipping = listingTypeConfig => {
   return listingTypeConfig?.defaultListingFields?.shipping !== false;
 };
 
+// TODO: update endpoint as needed
+export const requireListingFiles = listingTypeConfig => {
+  return listingTypeConfig?.defaultListingFields?.files === true;
```
Gnito, line 1100: There's one interesting nuance here. The other require* and display* helpers use `!== false` check. This is because the default value for undefined info is interpreted as "active". E.g. assume price is shown if price is not defined.

With this helper, you have used `=== true` and to me that sounds correct. If "files" is not defined, we assume that the files are not required for the particular listing type.

Long story short: I think this is correct, but it would be good to add code comment why this is behaving differently than the others.

Side note: if we want this to work the same way as the other helpers, we would need to make migration to every listing type assets across all the marketplaces (and add `files: false,` to every listing type).

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesForm.example.js
```diff
@@ -0,0 +1,17 @@
...
+    onSubmit: values => {
+      console.log('Submit EditListingFilesForm with values:', values);
+    },
+    saveActionMsg: 'Next',
+    updated: false,
+    updateInProgress: false,
+    disabled: false,
+    ready: false,
+  },
+  group: 'page:EditListingPage',
```
Gnito: Remember to check this. Currently, I don't see any option to upload files, and I'm not sure we can/want upload anything on styleguide. So, the question is if there's some way to mock the UI (without allowing any actual upload) - or should we just drop the example.

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesForm.test.js
```diff
@@ -0,0 +1,28 @@
...
+import { renderWithProviders as render, testingLibrary } from '../../../../util/testHelpers';
+
+import EditListingFilesForm from './EditListingFilesForm';
+
+const { screen } = testingLibrary;
+
+const noop = () => null;
+
+describe('EditListingFilesForm', () => {
+  it('renders with submit button enabled', () => {
```
Gnito: What about "Upload a file" button? Can we show that?

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesPanel.js
```diff
@@ -0,0 +1,165 @@
...
+    updatePageTitle: UpdatePageTitle,
+    intl,
+    hasPendingFileUploads,
+  } = props;
+
+  const config = useConfiguration();
+
+  const rootClass = rootClassName || css.root;
+  const classes = classNames(rootClass, className);
+  const isPublished = listing?.id && listing?.attributes?.state !== LISTING_STATE_DRAFT;
```
Gnito: This is quite old check from the time, when we only had draft listings and published listings.

Perhaps, we should change this to 

```suggestion
  const isDraft = listing?.id && listing?.attributes?.state === LISTING_STATE_DRAFT;
```

and swap the panelHeadingProps ternary on this file and filesVerifiedIfPublished on EditListingFilesForm.js

src/translations/en.json
```diff
@@ -357,6 +363,12 @@
   "EditListingWizard.default-booking.new.savePhotos": "Publish listing",
   "EditListingWizard.default-booking.new.savePricing": "Next",
   "EditListingWizard.default-booking.new.saveStyle": "Publish listing",
+  "EditListingWizard.default-download.new.saveDetails": "Next",
+  "EditListingWizard.default-download.new.saveLocation": "Next",
+  "EditListingWizard.default-download.new.saveFiles": "Next",
```
Gnito, line 372: Perhaps, alphabetical sorting is needed here

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesPanel.js
```diff
@@ -0,0 +1,165 @@
...
+
+  const allowFiles =
+    !config.accessControl.marketplace.fileUploadAndDownloadDisabled && !fileUploadsDisabled;
+  // TODO: update when config ready
+  const listingTypeHasFileAttachments = foundListingTypeConfig?.defaultListingFields?.files;
+  const showAttachFiles = listingTypeHasFileAttachments && allowFiles;
+
+  const onUploadFileToPanel = file => {
+    if (file) {
+      const tempId = `${Date.now()}-${Math.random()}`;
```
Gnito: Perhaps this could have some prefix (so that debugging is more understandable). 

```suggestion
      const tempId = `${file.name}-${Date.now()}`;
```

https://developer.mozilla.org/en-US/docs/Web/API/File/name

src/containers/EditListingPage/EditListingWizard/EditListingStylePanel/EditListingStyleForm.js
```diff
@@ -104,7 +105,7 @@ export const EditListingStyleForm = props => {
 
         const submitInProgress = updateInProgress;
         const submitReady = updated || ready;
-        const submitDisabled = submitInProgress;
+        const submitDisabled = disabled || submitInProgress;
```
Gnito, line 111: What about photos form?

Wizard can end with styles form or photos form => both need to handle the disabled submit the same way.

src/containers/EditListingPage/EditListingWizard/EditListingWizard.js
```diff
@@ -102,6 +106,13 @@ const tabsForListingType = (processName, listingTypeConfig) => {
     ['default-purchase']: [DETAILS, PRICING_AND_STOCK, ...deliveryMaybe, ...styleOrPhotosTab],
     ['default-negotiation']: [DETAILS, ...locationMaybe, ...pricingMaybe, ...styleOrPhotosTab],
     ['default-inquiry']: [DETAILS, ...locationMaybe, ...pricingMaybe, ...styleOrPhotosTab],
+    ['default-download']: [
+      DETAILS,
+      ...locationMaybe,
+      ...filesTab,
+      ...pricingMaybe,
+      ...styleOrPhotosTab,
+    ],
```
Gnito: Files tab is mandatory for this process. I think we should make it explicit here.


```suggestion
    ['default-download']: [
      DETAILS,
      ...locationMaybe,
      FILES,
      ...pricingMaybe,
      ...styleOrPhotosTab,
    ],
```

src/containers/EditListingPage/EditListingPage.duck.js
```diff
@@ -546,6 +571,225 @@ const fetchLoadDataExceptions = (dispatch, listing, search, firstDayOfWeek) => {
...
+  'EditListingPage/pollForFileVerification',
+  pollForFileVerificationPayloadCreator
+);
+
+// Backward compatible wrappers for the thunks
+export const pollForFileVerification = (fileId, tempId) => dispatch => {
+  return dispatch(pollForFileVerificationThunk({ fileId, tempId }));
+};
+
+// TODO: extract this from here and TransactionPage.duck.js to make a re-usable helper
```
Gnito: Highlighting this todo comment

Gnito: Perhaps, the inner composition could be moved to some helper function (e.g. inside fileHelpers.js 🤔)

I.e. this composeAsync and related functions functions:
```js
  const handleFileUpload = composeAsync(
    checkMetadataFn,
    createFileResourceFn,
    createFileUploadUrlFn,
    uploadFileToStorageFn,
    handleFileUploadSuccessFn
  );
```

src/containers/EditListingPage/EditListingPage.js
```diff
@@ -158,6 +170,13 @@ export const EditListingPageComponent = props => {
     stripeAccount,
     updateStripeAccountError,
     authScopes,
+    fileUploads,
+    fileUploadsDisabled,
+    hasPendingFileUploads,
+    allFilesUploadedAndVerified,
```
Gnito, line 177: JSDocs doesn't mention these

src/containers/EditListingPage/EditListingPage.duck.js
Gnito, line 910: As discussed over huddle, `state.fileUploads` should be updated with the file array found from the response (`action.payload.response.data`) as that's the current list of attached files on API.
and a similar change to `publishListingThunk.fulfilled` case too.

src/containers/EditListingPage/EditListingPage.duck.js
```diff
@@ -685,18 +969,59 @@ const editListingPageSlice = createSlice({
...
+              verificationInProgress: false,
+              error: null,
+              file,
+              sourceFile: null,
+              progress: null,
+              verificationStatus: file?.attributes?.state,
+              tempId: id,
+            };
+          }
+        });
```
Gnito: You might want to turn this into a helper function inside this file, because (I think) you need to use it also on _updateListingThunk.fulfilled_ and _publishListingThunk.fulfilled_

Although, _updateListingThunk.fulfilled_ and _publishListingThunk.fulfilled_ does not need to expect any just uploaded files with tempId => the list returned from the API call among listing entity is the single source of truth.

This however, might contain both. E.g. user can click the tab link on the page, which would then call loadData.

So, I'm not entirely sure the change I proposed on today's huddle was good (aka clear fileUploads on loadData).

Perhaps there should be extra if-statement around it.
`if (id ===  getState().EditListingPage.listingId?.uuid) {`
where the "id" is the id found from params (URL path params).

src/containers/EditListingPage/EditListingWizard/EditListingStylePanel/EditListingStyleForm.js
Amanda-Kauppinen: ✅ photos are also handled! I did them in separate commits

src/util/configHelpers.js
Amanda-Kauppinen: ✅

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesForm.example.js
Amanda-Kauppinen: ✅ removed

src/translations/en.json
Amanda-Kauppinen: ✅ (not showing as outdated, but this is an old view of the `en.json` file)

src/containers/EditListingPage/EditListingPage.js
Amanda-Kauppinen: ✅

src/containers/EditListingPage/EditListingPage.duck.js
Amanda-Kauppinen: ✅

src/containers/EditListingPage/EditListingPage.duck.js
```diff
@@ -546,6 +619,128 @@ const fetchLoadDataExceptions = (dispatch, listing, search, firstDayOfWeek) => {
...
+  { fileAttachmentId, isOwnFile },
+  { rejectWithValue, extra: sdk }
+) => {
+  if (!fileAttachmentId) {
+    throw new Error('Missing fileAttachmentId, cannot initiate download.');
+  }
+  // Request a temporary download URL from the SDK
+  const downLoadFn = isOwnFile
+    ? sdk.ownFileDownloads.create({ fileId: fileAttachmentId })
+    : sdk.fileDownloads.create({ fileAttachmentId });
```
lyyder: `sdk.onwFileDownloads` should take the file ID, as the param indicates, but here the value passed is `fileAttachmentId`. Is that wrong, or does this function receive file ID as the value of `fileAttachmentId` in case this is the user's own file? In that case the the param could possibly be named `fileOrAttachmentId` etc to avoid confusion.

lyyder: Thinking this further, should this always be invoked by the file's owner, so this would always use the `ownFileDownload`.

Amanda-Kauppinen: @lyyder good catch - this should be renamed. it passes `file.id.uuid` (happens in [FileUpload.js](https://github.com/sharetribe/web-template/blob/ccb59178665e7e2372aafb4005fed81d6a3d5079/src/components/FileUpload/FileUpload.js#L162))

Right now this is in the main branch's transaction messages flow, so I'll change the name from `fileAttachmenId` → `fileId` in both places. I think it's okay that it gets fixed when my branch eventually merges since functionally it works correctly but it is misleading since we have both `files` and `fileAttachments` - do you think it should be fixed sooner than that?

Amanda-Kauppinen: sorry! hadn't refreshed to see the second comment @lyyder - that makes sense to only use `ownFileDownload` here as opposed to transaction messages where both parties can see them. will update

src/containers/EditListingPage/EditListingWizard/EditListingFilesPanel/EditListingFilesPanel.js
```diff
@@ -0,0 +1,166 @@
...
+      };
+
+  const validListingTypes = config.listing.listingTypes;
+  const foundListingTypeConfig = validListingTypes.find(
+    conf => conf.listingType === listing?.attributes?.publicData?.listingType
+  );
+
+  const allowFiles =
+    !config.accessControl.marketplace.fileUploadAndDownloadDisabled && !fileUploadsDisabled;
+  // TODO: update when config ready
```
SariSaar: Is this still a todo?

Amanda-Kauppinen: oops, no 😄 I'll remove it, thanks!

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.js
```diff
@@ -281,6 +283,21 @@ export const EditListingPhotosForm = props => {
             <PublishListingError error={publishListingError} />
             <ShowListingsError error={showListingsError} />
 
+            {!allFilesUploadedAndVerified ? (
```
SariSaar: This is now shown also for listings in with non-downloads listing types – publishing the listing works, but the error message is unnecessary

Image

src/containers/EditListingPage/EditListingPage.duck.js
```diff
@@ -843,6 +1095,88 @@ const editListingPageSlice = createSlice({
...
+          file: null,
+          sourceFile: file, // Set source file so that other components can get the file name
+          progress: null,
+          verificationStatus: null,
+          tempId,
+        };
+      })
+      .addCase(uploadFileThunk.fulfilled, (state, action) => {
+        const { fileUpload, tempId } = action.payload;
+        state.fileUploads[tempId].file = fileUpload;
```
SariSaar: Another thing that I added in acceptance testing for message files – wrapping this in an if(state.fileUploads[tempId]) block prevents Sentry errors when e.g. a user removes the upload.

PR #861, opened 2026-06-15
Fix scrollToMessage and streamline file download behavior

SariSaar:
- Use the right message id to scroll to message
- Use the same shape of file id for own file download as for sent message download
- Add guards for fileAttachmentId before handling state.fileDownloads


Gnito (approved):
LGTM

PR #874, opened 2026-07-07
CreateUserWithIdp: split user data to 2 cookies

Gnito:
st-authinfo contains only ConfirmSignupForm related data.
The new cookie: st-idp-token contains idpToken and is httpOnly


lyyder (approved):
Looks good! One thing could have been to also store `idpId` in the distinct cookie. Not that it's secret, but it's also not to be modified by the user when creating an account and it would be nicely bundled with the token. But keeping the distinct cookie only for the token also communicates that it's something we don't want to leak out so fine like this also.

PR #889, opened 2026-07-29
Sort listing images

Amanda-Kauppinen:
Adds functionality for reordering images while creating/editing a listing. It uses [SortableJS](https://sortablejs.github.io/Sortable/) and allows rearranging via drag-and-drop or with keyboard commands with accessibility support. Images are added/removed with one API call.

### New translation keys:

```diff
+  "EditListingPhotosForm.coverImageBadge": "Cover",
+  "ListingImage.screenreader.currentPosition": "{index} of {count}",
+  "ListingImage.screenreader.dragToReorderImage": "Drag to reorder",
+  "ListingImage.screenreader.moveImageDown": "Move image down - current position {currentPosition}.",
+  "ListingImage.screenreader.moveImageDownDisabled": "Move image down, disabled — last image",
+  "ListingImage.screenreader.moveImageUp": "Move image up - current position {currentPosition}.",
+  "ListingImage.screenreader.moveImageUpDisabled": "Move image up, disabled — first image",
```

### Old translations with updated arguments:

```diff
- "EditListingPage.screenreader.removeImage": "Remove image",
+ "EditListingPage.screenreader.removeImage": "Remove image - current position {currentPosition}.",
```


src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.js
```diff
@@ -21,6 +22,41 @@ import css from './EditListingPhotosForm.module.css';
 
 const ACCEPT_IMAGES = 'image/*';
 
+// Split out of the main bundle - only fetched once this form actually mounts.
+const Sortable = loadable.lib(() => import(/* webpackChunkName: "sortablejs" */ 'sortablejs'));
+
+const SortableController = props => {
```
Gnito, line 36: Add code-comments here about what this is and how it's related to Sortable lib & images in FieldArray.

(I don't think we have this pattern anywhere in the codebase - so, it's good to have some extra comments next to it.)

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.module.css
```diff
@@ -12,14 +12,52 @@
...
+.sortableChosen > * {
+  transition: transform 0.2s ease-in, box-shadow 0.2s ease-in;
+  transform: scale(1.03);
+  box-shadow: rgba(0, 0, 0, 0.25) 0px 20px 40px;
+}
+
+.sortableDrag {
+  overflow: visible;
+  user-select: none;
+  opacity: 1 !important;
```
Gnito, line 66: Add code-comment here to explain why !important is needed.
(The Sortable.js lib sets `opacity: 0.8` through inline-styles and we want to overwrite that.)

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.module.css
```diff
@@ -12,14 +12,52 @@
...
+
+  @media (--viewportMedium) {
+    grid-template-columns: 1fr 1fr;
+  }
 }
 
 .thumbnail {
   width: 100%;
   background-color: var(--colorGrey50);
+  border-radius: var(--medium, 4px);
```
Gnito: Go through all the `--medium` variables. This is some sort of hallucination. There's no variable like that.
However, there's `--borderRadiusMedium`
https://github.com/sharetribe/web-template/blob/main/src/styles/marketplaceDefaults.css#L139

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/ListingImage.js
```diff
@@ -68,19 +146,118 @@ const ListingImage = props => {
...
+            onMoveDown={handleMoveDownClick}
+            disableMoveUp={disableMoveUp}
+            disableMoveDown={disableMoveDown}
+          />
+        ) : (
+          <DragHandleButton buttonRef={handleButtonRef} />
+        )}
+        <RemoveImageButton buttonRef={removeButtonRef} onClick={handleRemoveClick} />
+      </div>
+    );
```
Gnito: I think this should be turned into a component. 

Perhaps in a way where buttons are passed in as props.
I assume that this ListingImage component does not need to know about controls focus/blur handling. So, ControlButtons would own that code.

```jsx
<ControlButtons
  hideMoveControls={hideMoveControls || !canReorder}
  {/* other props if needed */}
  moveImageButtons={MoveImageButtons}
  dragHandleButton={DragHandleButton}
  removeImageButton={RemoveImageButton}
/>
```

We have used a pattern like this here and there.
E.g. 
- https://github.com/sharetribe/web-template/blob/21fcf8974434f1bb3d80173462ccd0c258694248/src/components/OrderPanel/OrderPanel.js#L531
- https://github.com/sharetribe/web-template/blob/21fcf8974434f1bb3d80173462ccd0c258694248/src/components/OrderPanel/BookingDatesForm/BookingDatesForm.js#L623

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/ListingImage.module.css
```diff
@@ -12,33 +12,90 @@
...
 
-.removeImage {
+.controlsGroup {
   position: absolute;
-  top: 0;
-  right: 0;
+  top: 8px;
+  right: 8px;
+  display: flex;
+  z-index: 1;
```
Gnito: I think this is wrong.
(I earlier suggested z-index:1 but only for .controlButton.

I.e. next to 
```css
  &:hover svg {
    fill: var(--colorWhite);
    stroke: var(--colorWhite);
  }

```

Add 

```css
  &:focus {
    z-index: 1;
  }
```

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/ListingImage.module.css
```diff
@@ -12,33 +12,90 @@
...
+
+.coverBadge {
+  position: absolute;
+  bottom: 8px;
+  left: 8px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  gap: 8px;
+  padding: var(--spacing-sm, 6px) var(--spacing-md, 8px);
```
Gnito: These `--spacing-` variables are hallucination. We don't have those.

But we have these: https://github.com/sharetribe/web-template/blob/21fcf8974434f1bb3d80173462ccd0c258694248/src/styles/marketplaceDefaults.css#L107-L108

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.js
```diff
@@ -229,24 +301,68 @@ export const EditListingPhotosForm = props => {
...
+                    }
+                    fields.move(fromIndex, toIndex);
+                    return true;
+                  };
+                  moveImageRef.current = moveImage;
+
+                  return (
+                    <div className={css.sortableImages} ref={sortableContainerRef}>
+                      {fields.map((name, index) => (
+                        <FieldListingImage
```
Gnito, line 396: You could pass a prop like this one to FieldListingImage:
```jsx
currentPositionMessage={intl.formatMessaga({ id: 'ListingImage.screenreader.currentPosition' }, { index: i + 1, count: fields.length })}
```

add `"ListingImage.screenreader.currentPosition": "{index}/{count}"` to en.json
or "{index} of {count}" as I'm not sure how understandable a message like "three slash ten" would be.

Then the currentPositionMessage could be passed to up & down translations:
```json
"ListingImage.screenreader.moveImageDown": "Move image down - current position {currentPosition}.",
"ListingImage.screenreader.moveImageUp": "Move image up - current position {currentPosition}.",
```

Note: we should ask Katri to check all these messages given for screenreader.

src/containers/EditListingPage/EditListingWizard/EditListingPhotosPanel/EditListingPhotosForm.module.css
Amanda-Kauppinen: Ahh. This isn't a claude failing after all. Well I mean slightly for not checking if that variable exists. But it's something I had put into the spec because I copied the styles over from figma and advised it to follow those rules 😅

Gnito (approved):
LGTM

PR #890, opened 2026-07-30
Harden Mapbox loading

SariSaar:
Fix an issue where a full page load might leave the Mapbox token unassigned: on a full page load (SSR + hydration) where the browser hasn't finished downloading/executing the external `mapbox-gl.js` script in `includeScripts.js` by the exact moment `IncludeScripts` first renders client-side, `window.mapboxgl.accessToken` never gets set for the rest of that page's lifetime.

The associated error that this fix is addressing is `TypeError: Cannot read properties of undefined (reading 'geocoding')`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ way to update this template, but currently, we follow a pattern:
 
 ## Upcoming version 2026-XX-XX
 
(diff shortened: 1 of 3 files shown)


Gnito (approved):
LGTM

PR #896, opened 2026-08-06
Availability in 15 min increments

SariSaar:


src/util/dates.js
```diff
@@ -821,8 +821,8 @@ export const getBoundaries = (startTime, endTime, unitCount, timeUnit, timeZone,
  *
  * @returns {Array} an array of objects with keys timestamp and timeOfDay.
  */
-export const getSharpHours = (startTime, endTime, timeZone, intl) => {
-  return getBoundaries(startTime, endTime, 1, 'hour', timeZone, intl);
+export const getSharpHours = (startTime, endTime, timeZone, intl, timeUnit = 'hour') => {
+  return getBoundaries(startTime, endTime, 1, timeUnit, timeZone, intl);
```
Gnito: Don't repurpose a function that's named to do a single job.

Instead of overloading a function called "getSharpHours" just use the underlying one (getBoundaries) - or create differently named function.

src/util/dates.js
```diff
@@ -852,8 +852,8 @@ export const getSharpHours = (startTime, endTime, timeZone, intl) => {
  *
  * @returns {Array} an array of objects with keys timestamp and timeOfDay.
  */
-export const getStartHours = (startTime, endTime, timeZone, intl) => {
-  const hours = getSharpHours(startTime, endTime, timeZone, intl);
+export const getStartHours = (startTime, endTime, timeZone, intl, timeUnit = 'hour') => {
+  const hours = getSharpHours(startTime, endTime, timeZone, intl, timeUnit);
```
Gnito: Create another function instead of making getStartHours to return something that is not an hour.

src/util/dates.js
```diff
@@ -884,8 +884,8 @@ export const getStartHours = (startTime, endTime, timeZone, intl) => {
  *
  * @returns {Array} an array of objects with keys timestamp and timeOfDay.
  */
-export const getEndHours = (startTime, endTime, timeZone, intl) => {
-  const hours = getSharpHours(startTime, endTime, timeZone, intl);
+export const getEndHours = (startTime, endTime, timeZone, intl, timeUnit = 'hour') => {
+  const hours = getSharpHours(startTime, endTime, timeZone, intl, timeUnit);
```
Gnito: Create another function instead of making getEndHours to return something that is not an hour.

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityPlanForm/AvailabilityPlanEntries.js
```diff
@@ -35,10 +61,11 @@ const ALL_END_HOURS = HOURS.map((v, i) => printHourStrings(i + 1));
  * @returns localized time format (e.g. '9:00 AM')
  */
 const localizedHourStrings = (hour24, intl) => {
-  const hour = Number.parseInt(hour24.split(':')[0]);
+  const { hour, minutes } = getParsedHourMinutes(hour24);
```
Gnito: Is this function anymore "localizedHourStrings"?

I think this needs to be either renamed or there should be another function.
Perhaps, **_localizedTimeStrings_**

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityExceptionForm/ExceptionDateTimeRange.js
```diff
@@ -124,8 +137,8 @@ const getAvailableEndTimes = ({
     : dayAfterSelectedEndDate;
 
   const selectableHours = isSingleDayRange
-    ? getEndHours(limitStart, limitEnd, timeZone, intl)
-    : getSharpHours(limitStart, limitEnd, timeZone, intl);
+    ? getEndHours(limitStart, limitEnd, timeZone, intl, timeUnit)
+    : getSharpHours(limitStart, limitEnd, timeZone, intl, timeUnit);
```
Gnito: This naming is confusing. Use functions that don't reference 'hour'

Perhaps something like this:
```diff
-   const selectableHours = isSingleDayRange
-     ? getEndHours(limitStart, limitEnd, timeZone, intl)
-     : getSharpHours(limitStart, limitEnd, timeZone, intl);
+   const selectableEntryBoundaries = isSingleDayRange
+     ? getEndBoundaries(limitStart, limitEnd, timeUnit, timeZone, intl)
+     : getBoundaries(limitStart, limitEnd, timeUnit, timeZone, intl);
```

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityExceptionForm/ExceptionDateTimeRange.js
```diff
@@ -84,7 +96,7 @@ const getAvailableStartTimes = ({ selectedStartDate, availableSlots, intl, timeZ
     // Otherwise use the end of the timeslot.
     const endLimit = isDateSameOrAfter(end, nextDay) ? nextDay : end;
 
-    const hours = getStartHours(startLimit, endLimit, timeZone, intl);
+    const hours = getStartHours(startLimit, endLimit, timeZone, intl, timeUnit);
```
Gnito: This is not about hours anymore. I think you need to change the names

src/util/dates.js
SariSaar: Separated into two functions now

src/util/dates.js
SariSaar: Done!

src/util/dates.js
```diff
@@ -889,6 +889,118 @@ export const getEndHours = (startTime, endTime, timeZone, intl) => {
...
+ * @param {Date} End point of available time window.
+ * @param {String} timezone name. It should represent IANA timezone key.
+ * @param {Object} intl containing formatting options for Intl.DateTimeFormat.
+ * @param {String} timeUnit scope. e.g. 'hour', 'quarterHour'. Defaults to 'hour'.
+ *
+ * @returns {Array} an array of objects with keys timestamp and timeOfDay.
+ */
+export const getSharpBoundaries = (startTime, endTime, timeZone, intl, timeUnit = 'hour') => {
+  return getBoundaries(startTime, endTime, 1, timeUnit, timeZone, intl);
+};
```
Gnito, line 929: So, this `getSharpBoundaries` is essentially just fixing unitCount to be exactly 1, and then calls more generic `getBoundaries`.

I wonder if we could figure out a better name for it.
- getAllBoundaries
- getUnitBoundaries
- ?

src/components/FieldSelectPopup/FieldSelectPopup.helpers.js
```diff
@@ -0,0 +1,75 @@
...
+  let node = element?.parentElement;
+  while (node && node !== document.body) {
+    const { overflowY } = window.getComputedStyle(node);
+    if (overflowY === 'auto' || overflowY === 'scroll' || overflowY === 'hidden') {
+      return node.getBoundingClientRect();
+    }
+    node = node.parentElement;
+  }
+  return { top: 0, bottom: window.innerHeight };
+};
```
Gnito, line 19: I felt that this JSDoc description was pretty difficult to understand. So, I asked AI to improve it.

The suggestion was something like this:

```js
/**
 * Returns the vertical viewport bounds that a floating UI (menu, tooltip, popover)
 * should stay inside when positioned relative to `element`.
 *
 * Walks up from `element`'s parent looking for the nearest ancestor that can clip
 * content vertically (`overflow-y: auto | scroll | hidden`). That ancestor's
 * `getBoundingClientRect()` is returned.
 *
 * If none is found before `<body>`, returns the visible viewport:
 * `{ top: 0, bottom: window.innerHeight }`.
 *
 * Coordinates are viewport-relative (same space as `getBoundingClientRect()`).
 *
 * @param {HTMLElement} element - Element whose ancestors are searched (typically the trigger)
 * @returns {{ top: number, bottom: number }} Vertical clip/scroll bounds in viewport pixels
 */
```

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -0,0 +1,388 @@
...
+  // Only moves the highlight; it never commits a value on its own.
+  const handleTypeahead = event => {
+    if (event.key.length !== 1 || event.ctrlKey || event.altKey || event.metaKey) {
+      return;
+    }
+    window.clearTimeout(typeaheadResetTimeoutRef.current);
+    typeaheadQueryRef.current += event.key;
+    typeaheadResetTimeoutRef.current = window.setTimeout(() => {
+      typeaheadQueryRef.current = '';
+    }, 500);
```
Gnito, line 161: This is somewhat buried "magic number". 

Create a constant instead (after imports). Perhaps something like:

```js
// The delay after which the accumulated query is cleared when typing pauses
const TYPEAHEAD_RESET_MS = 500;
```

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -0,0 +1,388 @@
...
+  // The keyboard-highlighted option index, independent of `input.value` (the committed
+  // selection). -1 means "no highlight" (always true while closed). Drives both the visual
+  // `.popupOptionHighlighted` class and `aria-activedescendant`: DOM focus itself never leaves the
+  // trigger button, so closing on blur (tabbing away) can coexist with clicking an option without
+  // one breaking the other.
+  const [highlightedIndex, setHighlightedIndex] = useState(-1);
+  // Set synchronously (not via state, see `handleKeyUp`'s comment below) whenever an Escape
+  // keydown closes the popup, so the matching keyup (a separate DOM event this component can't
+  // otherwise tie back to that keydown) knows to stop itself from also closing an ancestor Modal.
+  const suppressNextEscapeKeyUpRef = useRef(false);
```
Gnito, line 49: I like that there's an explanation alread here - but this particular formulation feels confusing.

I think the explanation on `case 'Escape':` was more clear:

          // keydown and keyup are independent DOM events for the same key press, so
          // preventDefault here has no effect on the keyup that follows. Without suppressing that
          // keyup too, it bubbles to a wrapping Modal's own Escape listener
          // (`Modal.js`'s `handleBodyKeyUp`) and closes the Modal too.

So, perhaps taking something from that description and add it here too.

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -0,0 +1,388 @@
...
+    const selectedIndex = options.findIndex(o => o.value === input.value && !o.disabled);
+    setHighlightedIndex(selectedIndex !== -1 ? selectedIndex : enabledIndices[0] ?? -1);
+  }, [isOpen]);
+
+  // Preserves the "opens scrolled to the current value" behavior a native <select> gives for
+  // free, now that the list can be taller than fits on screen (see .popupList's max-height
+  // below), and keeps the keyboard-highlighted option in view the same way while navigating.
+  // Sets `list.scrollTop` directly instead of `node.scrollIntoView({ block: 'center' })`, which
+  // also scrolled a wrapping Modal's own scroll layer, causing a visible jump. Clamped to the
+  // list's own scrollable range so an option near the very top/bottom doesn't overscroll.
```
Gnito, line 97: State the effect of the code first. E.g.

```js
// Keep the highlighted option roughly centered in the list. That covers both opening on the
// current value (like a native <select>) and keyboard navigation afterward. The list can
// outgrow the viewport via .popupList's max-height.
//
// Set `list.scrollTop` directly instead of `node.scrollIntoView({ block: 'center' })`: the latter
// also scrolls a wrapping Modal and causes a visible jump. Clamp to the list's scroll range so
// options near the top/bottom don't overscroll.
```

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -0,0 +1,388 @@
...
+  const classes = classNames(rootClassName || css.root, className);
+  const triggerClasses = classNames(css.trigger, {
+    [selectClassName]: selectClassName,
+    [css.triggerError]: hasError,
+  });
+
+  // ARIA ids. `labelId` only exists when `label` is actually rendered (it's an optional prop:
+  // AvailabilityPlanEntries.js, for instance, relies on its own external <label> instead), so
+  // `aria-labelledby` is only wired up when there's something for it to point at.
+  const labelId = label ? `${id}-label` : undefined;
```
Gnito, line 278: What is the string currently used for entry in the default schedule?
((there's probably some bug)

That being said, I think it should be possible to add separate aria-label for these start time & end time buttons.

Something like: 
- "Availability entry for Monday. Selected start time is 10: 45 am. Selected end time is 15:00 am"
- "Invalid availability entry for Monday. End time is missing."

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityPlanForm/AvailabilityPlanEntries.js
```diff
@@ -136,65 +139,71 @@ const filterEndHours = (availableEndHours, entries, index) => {
...
-      const hoursBetween = Array(endHour - startHour)
-        .fill()
-        .map((v, i) => printHourStrings(startHour + i + boundaryDiff));
-
-      return allHours.concat(hoursBetween);
-    }
-
-    return allHours;
-  }, []);
+const TimeSelectField = props => {
```
Gnito, line 177: Don't change the FinalForm's field naming pattern here: start with Field*

```suggestion
const FieldSelectTime = props => {
```

That way we find only form fields with typeahead search of "`<Field`"

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityExceptionForm/ExceptionDateTimeRange.js
```diff
@@ -301,6 +346,37 @@ const onExceptionEndDateChange = (value, availableSlots, props) => {
...
+ * FieldSelectPopup (losing its open/highlighted-option state) instead of updating it in place.
+ *
+ * @component
+ * @param {Object} props
+ * @param {Boolean} props.useIncrementalBoundaries whether to render FieldSelectPopup (true) or FieldSelect (false)
+ * @param {Array<{value: string, label: ReactNode, disabled: boolean}>} props.options
+ * @param {...*} rest forwarded to whichever component is rendered
+ * @returns {JSX.Element}
+ */
+const TimeSelectField = props => {
```
Gnito, line 365:
```suggestion
const FieldSelectTime = props => {
```

src/containers/EditListingPage/EditListingWizard/EditListingAvailabilityPanel/EditListingAvailabilityPlanForm/AvailabilityPlanEntries.js
Gnito: Or if the point is that this is not real final form "field" but a wrapper: then perhaps just 
```js
const SelectTime = props => {
```

Edit. that being said: FieldSelectTime would be in sync with FieldDateAndTimeInput component, which is also just a wrapper.

src/components/FieldSelectPopup/FieldSelectPopup.js
Gnito: Ah. After re-deploy, it was a bit easier to understand how this works. => This creates a hidden label for accessibility purposes (and combines 2 info together)

`aria-labelledby="mon[0].startTime-label mon[0].startTime-value"`
=> "Monday start time" & "12:00 AM"

Screenshot 2026-09-07 at 17 45 16

Perhaps this pattern could be documented as a code-comment.

To my understanding, this also allows sentences like:
_"Availability entry for Monday, starting at 12:00 AM"_

Gnito: The problematic scenario is invalid aka unselected start/end.

Then this current setup says: "Monday start time Start"

src/components/FieldSelectPopup/FieldSelectPopup.module.css
```diff
@@ -0,0 +1,161 @@
...
+  position: relative;
+}
+
+/* Stays a plain in-flow child. Clipping by a sibling row's own stacking context
+ * (AvailabilityPlanEntries.module.css's `.timeRangeRow`: `position: relative` + `z-index: 1`) is
+ * fixed by the caller raising that row's own z-index above its siblings while this popup is open
+ * (see FieldSelectPopup.js's `onToggleActive` prop), not by escaping the DOM subtree via a portal:
+ * a portal's reach also means competing with unrelated ancestors' z-index (e.g. a wrapping
+ * Modal's), which an in-flow popup never has to do. `--zIndexPopup` per marketplaceDefaults.css's
+ * convention for small on-page popups; only needs to out-rank this popup's own local siblings. */
```
Gnito, line 99: Perhaps this comment could be rephrased to be more human friendly.

src/components/FieldSelectPopup/FieldSelectPopup.helpers.js
SariSaar: I did another round of rewrites for all comments now

src/components/FieldSelectPopup/FieldSelectPopup.js
SariSaar: This is now how it looks. I also added the logic as a comment.
Screenshot 2026-09-18 at 11 36 48
Screenshot 2026-09-18 at 11 37 28

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -9,7 +9,6 @@ import { OutsideClickHandler, ValidationError, HelpText } from '../../components
 import { getScrollBoundary, moveHighlight, findTypeaheadMatch } from './FieldSelectPopup.helpers';
 import css from './FieldSelectPopup.module.css';
 
-// The delay after which the accumulated query is cleared when typing pauses
```
Gnito, line 12: I'd keep this one

src/components/FieldSelectPopup/FieldSelectPopup.js
```diff
@@ -69,19 +65,16 @@ const FieldSelectPopupComponent = props => {
   };
 
   const { invalid, touched, error } = meta;
-
-  // Error message and input error styles are only shown if the
-  // field has been touched and the validation has failed.
```
Gnito, line 74: I assume this is just accidental deletion. The original helps to understand why the existence of the error object is not enough for showing the error message (which might happen already with the initial render).

PR #904, opened 2026-08-12
Refactor checkout

Gnito:
This PR refactors CheckoutPage heavily and moves some payment method related code patterns close to transaction process as transaction process is the source of truth what payment methods it supports.

**If you are taking updates from upstream:**

- **Custom `transactionProcess*.js` files must export** :
  - `supportedPayments`
    - This is defines what kind of payments the process supports. Potentially something like this:
      ```js
      export const supportedPayments = {
        stripe: {
          ...stripePaymentMethodInfo(PAYMENT_METHOD_CARD),
          ...stripePaymentMethodInfo(PAYMENT_METHOD_IDEAL),
        },
      };
      ```
  - Move transitions needed on the checkout page to the process files
    - This way they can be named more freely
    - `getCheckoutPaymentTransitions({ paymentProcessor, paymentMethod, state, ... })`
    - returns `{ requestPaymentTransition, confirmPaymentTransition }`
  - `hasPaymentExpired(tx)` 
    - Moved out of checkout helpers; each process owns this check
- **Rename:** `SavedCardDetails` → `PaymentMethodPicker`
- **Card checkout behavior is unchanged by default**
  - Built-in processes still offer card only; the new return route / push paths stay dormant until a process declares a push method.

If you customized checkout, note the overlapping concepts that have used "payment method" internally:
 - checkoutPaymentMethod (new): catalog id from paymentMethods.js ('card', 'ideal', …)
 - cardPaymentMode (renamed): card UX ('defaultCard' | 'replaceCard' | 'onetimeCardPayment')
 - orderParams.paymentMethod: saved Stripe payment method id ('pm_…'), when using a saved default card

In addition, the purpose of this PR is also to make it easier to add support for additional payment methods.
- **This does not enable any push methods!** 
  - This only drafts a UI flow for redirect (“push”) payments that confirm on an external site.
- Helpers and a rough `CheckoutPageRedirectReturn` component (and its route) exist, but are not production-ready.
- Server allowlist for `paymentMethodTypes` is empty by default. 
  - It is a mandatory parameter for push payments
  - Update [`pushPaymentMethodValidation.js`](server/api-util/pushPaymentMethodValidation.js) if you take a push method into use.
- Full push support still needs careful planning: transaction process changes (new transitions / push Stripe actions) and a way to handle Stripe webhooks/events when the customer does not return from the 3rd party site.

See [`src/transactions/README.md`](src/transactions/README.md) and [`CheckoutPageRedirectReturn/README.md`](src/containers/CheckoutPage/CheckoutPageRedirectReturn/README.md).


src/containers/CheckoutPage/CheckoutPageRedirectReturn/README.md
```diff
@@ -0,0 +1,173 @@
...
+It is mounted by [CheckoutPage](../CheckoutPage.js) when the route uses
+`extraProps: { mode: 'return-after-redirect' }`
+(`[CheckoutRedirectReturnPage](../../../routing/routeConfiguration.js)` →
+`/l/:slug/:id/checkout/return`).
+
+## Not in active use by default
+
+**Built-in template processes only expose card (**`paymentDirection: 'pull'`**).** This directory,
+the `/checkout/return` route, and the push branch in checkout submit exist so customizers can enable
+redirect payment methods without reinventing the return-page UX — but **stock checkout never
```
SariSaar:
```suggestion
redirect payment methods without reinventing the return-page UX — but **default checkout never
```

I would reword away from "stock checkout" since stock has a specific meaning in the template. I had to read this twice to realize that it's not talking about stock vs booking checkout but instead about a default/built-in checkout.

src/containers/CheckoutPage/CheckoutPageRedirectReturn/README.md
```diff
@@ -0,0 +1,173 @@
...
+run in one submit chain on CheckoutPage.
+
+## Enabling Stripe push payment methods
+
+Follow the checklist in [transactions/README.md](../../../transactions/README.md) (pull vs push,
+process module contract). In short:
+
+### 1. Backend transaction process
+
+- Use Sharetribe actions appropriate for push (e.g. `:action/stripe-create-payment-intent-push`).
```
SariSaar:
```suggestion
- Use `:action/stripe-create-payment-intent-push` to create the payment intent.
```
There are no other push specific actions, and the confirm-and-capture logic is mentioned below, so I'd suggest simplifying this.

PR #928, opened 2026-09-15
Fix/borrower reshop link

amalyb:


Gnito:
This seems like an accidental PR. Closing

PR #929, opened 2026-09-19
user.duck / fetchCurrentUserHasListings: Don't read currentUser from the store.

jayenashar:
`fetchCurrentUserHasListingsPayloadCreator` reads `getState().user.currentUser` and resolves `{ hasListings: false }` when it is null. Since the Redux Toolkit rewrite of this duck, the currentUser only reaches the store when `fetchCurrentUserThunk` FULFILS, and `fetchCurrentUser` dispatches this fetch from inside its own `.then`, which runs before that. So on every cold page load the guard sees null and answers "no listings" without ever calling the API.

The guard is redundant as well as wrong: the only caller dispatches this inside `fetchCurrentUser`'s own `isUserAuthorized` branch, so there is a signed-in user, and `sdk.ownListings.query` is scoped to the caller's token and needs no user id. It was neither redundant nor wrong before the rewrite, where the query passed `authorId: currentUser.id.uuid`.

### What it changes on screen

`currentUserHasListings` is false for everyone on the page they land on, so:

- `ModalMissingInformation` skips the payout-details reminder for a provider who has listings and has not given their payout details, and skips the email-verification reminder for a provider with listings and no orders (`hasListingsOrOrders`).
- `Topbar` sends a user with both roles to `/inbox/orders` instead of `/inbox/sales`.

Both come right only after a later `fetchCurrentUser` runs with the user already in the store, which takes two in-app navigations. Someone who arrives, reads one page and leaves is never prompted.

### Reproducing it

Signed in as a provider with a published listing and no payout details, on a deployed marketplace:

1. Load any non-whitelisted page with a full page load. No payout reminder, and the Inbox link reads `/inbox/orders`. The request `own_listings/query?states=published&page=1&perPage=1` is never sent.
2. Click through to another page. The request goes out, but the modal check has already run.
3. Click once more. The payout reminder opens and the Inbox link reads `/inbox/sales`.

With this change, step 1 shows both.

The sibling `fetchCurrentUserOwnListingsCount` in a customization carried the same guard with the same result, which is how this one came to light.


Gnito:
Thank you for the clear bug report and pull request! I ended up extending the fix to fetchCurrentUserHasOrders too.
https://github.com/sharetribe/web-template/pull/930

